PR #94, opened 2015-03-13
Explicitly mention the need to create the cache table in the docs

drnlm:
We should mention it, since the schedule breaks without it.


hodgestar:
:+1:

PR #95, opened 2015-04-22
Update time selection helper to match Django's current behaviour

drnlm:
Django's DateTimeShortcuts changed after I wrote this helper, so it currently just inserts "NaN:NaN" as the selected time. 

This updates it to work correctly with Django 1.7, but it will break with Django 1.6. This is annoying but I don't think it's worth the bother of trying to support both here.


hodgestar:
:+1:

PR #96, opened 2015-04-22
Add field so we can create multiple slots for the schedule easily

drnlm:
This adds an "additional slots" field whn creating slots, so a number of slots of the same length can be created at once. 

Hopefully this will make creating the scaffolding for the schedule a bit simpler.

### DIFF
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -190,18 +190,22 @@ class ScheduleItemAdmin(admin.ModelAdmin):
 
 class SlotAdminForm(forms.ModelForm):
 
-    additional = forms.IntegerField(min_value=0, max_value=30, required=False,
-                                    label="Additional slots",
-                                    help_text="Create this number of "
-                                              "additional slots following"
-                                              "this one")
-
     class Meta:
         model = Slot
         fields = ('name', 'previous_slot', 'day', 'start_time', 'end_time')
 
     class Media:
         js = ('js/scheduledatetime.js',)
+
+
+class SlotAdminAddForm(SlotAdminForm):
+
+    # Additional field added for creating multiple slots at once
+    additional = forms.IntegerField(min_value=0, max_value=30, required=False,
+                                    label="Additional slots",
+                                    help_text="Create this number of "
+                                              "additional slots following"
+                                              "this one")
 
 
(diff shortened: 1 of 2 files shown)


wafer/schedule/admin.py
```diff
@@ -187,6 +189,13 @@ def changelist_view(self, request, extra_context=None):
 
 
 class SlotAdminForm(forms.ModelForm):
+
+    additional = forms.IntegerField(min_value=0, max_value=30, required=False,
```
hodgestar: Will this show up when a slot is being edited? If so, is there any easy way to hide it?

hodgestar:
Left one question. Looks good. Any chance we can get a test for `save_model`?

wafer/schedule/admin.py
drnlm: Good point - it will currently show up on the edit option.

It should be possible to hide it by tweaking SlotAdmin.get_form - I'll need to think a bit about how best to implement that.

drnlm:
Test case added. I think this addresses all the concernes.

wafer/schedule/admin.py
```diff
@@ -195,10 +201,20 @@ class Media:
         js = ('js/scheduledatetime.js',)
 
 
+class SlotAdminAddForm(SlotAdminForm):
+
+    # Additional field added for creating multiple slots at once
+    additional = forms.IntegerField(min_value=0, max_value=30, required=False,
+                                    label="Additional slots",
```
hodgestar: Should we be `_(...)` or lazytexting the label and help_text?

wafer/schedule/tests/test_admin.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertEqual(slot, slot2)
+
+    def test_save_model_change_slot(self):
+        """Test save_model changing a slot"""
+        slot = Slot(day=self.day, start_time=D.time(11, 0, 0),
+                    end_time=D.time(12, 30, 0))
+        # end_time is chosen as 12:30 so it stays valid through all the
+        # subsequent fiddling
+        slot.save()
+        # check that it's not saved in the database
```
hodgestar: s/not//

hodgestar:
Left two very small comments. Otherwise looks good.

PR #101, opened 2015-07-29
Bugs/document media root writable

drnlm:
This adds some reminders that MEDIA_ROOT's permissions must be set correctly, since we often forget to do that.


hodgestar:
:+1:

Issue #104, opened 2015-08-12
Update Wafer to work with Django 1.8.


hodgestar:

hodgestar:
Addressed by #105.

PR #105, opened 2015-08-12
Add Django 1.8 support.

hodgestar:
Starts to address #104.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,14 @@ python:
 env:
    - DJANGO_VERSION=1.6
    - DJANGO_VERSION=1.7
-# Django 1.7 doesn't support python 2.6, so explicitly exclude that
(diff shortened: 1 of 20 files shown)


drnlm:
:+1:

hodgestar:
I think there are still some templates that need `{% load i18n %}` added.

drnlm:
That's almost certainly the case, although I don't think that prevents merging this.

We should extend the test suite so we touch each template we provide anyway, which should help catch these, but that should be a separate thing from the Django 1.8 work.

hodgestar:
The issue at the moment is that deploying this will break whenever a
template that doesn't have the `{% load i18n %}` is loaded.

drnlm:
I've added enough load tags that static site generation works, which cover all the major parts of the site.

hodgestar:
Thanks!

:+1:

Issue #103, opened 2015-08-12
Protect against long ticket type names from Quicket.


hodgestar:
Currently if a ticket type name is very long, we fail with:
```
File "/home/pyconza/pyconza2015/ve/lib/python2.7/site-packages/wafer/tickets/views.py", line 82, in import_ticket
    type_, created = TicketType.objects.get_or_create(name=ticket_type)
...
File "/home/pyconza/pyconza2015/ve/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
DataError: value too long for type character varying(32)
```

hodgestar:
Addressed by #106.

PR #106, opened 2015-08-12
Truncate long ticket types and add tests.

hodgestar:
Addresses #103.


hodgestar:
Ready for review.

drnlm:
We only show the type in the admin interface and related management queries, right? If that's true, not logging is fine - it was mainly a thought for helping identify the cause of truncated tickets.

Increasing the size of the field does seems a good idea for the future.

hodgestar:
TicketType.name length increased to 255 characters. Ready for re-review.

drnlm:
Looks fine to me

:+1:

Issue #102, opened 2015-07-30
Edit button doesn't work for main /index page


drnlm:
The edit button is shown for the main /index page, but does nothing.

I haven't dug into this yet, but I'm pretty sure it's because of the current url patterns for pages, which mean we're not going down the path where we handle the edit option for the main index page.

drnlm:
Closed by #109

PR #109, opened 2015-08-13
Pass query string when redirecting to page views.

hodgestar:
Addresses #102.

### DIFF
--- a/wafer/pages/tests/test_pages.py
+++ b/wafer/pages/tests/test_pages.py
@@ -1,0 +1,90 @@
+# Simple test of the edit logic around pages
+
+from django.test import Client, TestCase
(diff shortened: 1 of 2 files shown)


hodgestar:
We should probably add tests.

drnlm:
This test case correctly fails on master and passes with Django 1.7 on the fixed branch for me.

It fails on Django 1.8 for me due to what looks like a Django 1.8 change - I haven't investigated how to fix that correctly yet.

drnlm:
Passes with Django 1.8 when #111 is merged into the mix.

hodgestar:
:+1: once #111 is merged in an Travis builds pass.

PR #111, opened 2015-08-14
Remove unneeded field specifier from EditPage form to make Django 1.8 happy

### LINKED ISSUE #110: Wafer page editing fails on Django 1.8 with ImproperlyConfigured: error
As seen on Travis, and confirmed locally, attempting to edit a page bombs out, ending with

"Specifying both 'fields' and 'form_class' is not permitted."
ImproperlyConfigured: Specifying both 'fields' and 'form_class' is not permitted.

drnlm:
The fields attribute in EditPage makes Django 1.8 unhappy, and doesn't seem to be required, so we drop it.

Closes #110

### DIFF
--- a/wafer/pages/views.py
+++ b/wafer/pages/views.py
@@ -15,7 +15,6 @@ class EditPage(UpdateView):
     template_name = 'wafer.pages/page_form.html'
     model = Page
     form_class = PageForm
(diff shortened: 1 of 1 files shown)


hodgestar:
:+1: if Travis builds pass.

PR #112, opened 2015-08-14
Bug fix more missing load i18n tags

drnlm:
This adds {% load i18n %} tags to a bunch of templates that need them

Every easily accesible page, including those on the admin site, works with these fixes.


hodgestar:
:+1:

PR #115, opened 2015-09-08
Issue 114 feature pentabarf xml support

drnlm:
Currently this produces a very basic view of the schedule - it doesn't include the talk abstract and other interesting details yet, but I think it's a reasonable starting point.

It passes my testing with giggity and confclerk. I've thrown some moderately realistic schedules at them now without anything blowing up, but there are probably corner cases we'll still uncover in due course.

### DIFF
--- a/wafer/schedule/models.py
+++ b/wafer/schedule/models.py
@@ -91,14 +91,16 @@ class Slot(models.Model):
         return self.start_time
 
     def get_duration(self):
+        """Return the duration of the slot as hours and minutes.
+
+           Used for the pentabarf export, which needs it in this format."""
         start = datetime.datetime.combine(self.get_day().date,
                                           self.get_start_time())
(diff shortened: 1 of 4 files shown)


wafer/schedule/models.py
```diff
@@ -88,6 +90,17 @@ def get_start_time(self):
...
 
+    def get_duration(self):
+        start = datetime.datetime.combine(self.get_day().date,
+                                          self.get_start_time())
+        end = datetime.datetime.combine(self.get_day().date,
+                                        self.end_time)
+        duration = (end - start).total_seconds()
+        result = {}
+        result['hours'] = duration // 3600
+        result['minutes'] = (duration - (3600 * result['hours'])) // 60
```
stefanor: Simplify:
```
result['hours'], result['minutes'] = divmod(duration // 60, 60)
```

PR #120, opened 2015-09-24
Use linebreak filter on bio so it's somewhat formattable

drnlm:
Currently, we don't show line breaks when showing the profile biogrpahy, which makes formatting this hard.

Arguably, this should be a markdown field, but this is at least is an improvement which doesn't require changing the database model.


hodgestar:
:+1:

hodgestar:
P.S. Markdown is a bit tricky for content submitted by non-staff since there are security implications.

Issue #119, opened 2015-09-21
pentabarf.xml missing stuff


CarlFK:
https://za.pycon.org/schedule/pentabarf.xml

```
<conf_url>/talks/14/</conf_url>  
```
should be the full url

missing:
twitter_id
released  (There is no field in the model, so just hard code it to True)

CarlFK:
no contact for 2nd speaker:
```
<event id="31">
<persons>
<person contact="..." id="53">Tsatsi Mahase</person>
<person id="54">Bukiwe Tsoeu</person>
```

drnlm:
The behaviour of <conf_url></conf_url> matches that of summit's pentabarf export.

Neither Giggity nor confclerk appear to use it, so we can change it without breaking those, but I'm still reluctanct to deviate too much from how other conferences produce pentabarf files.

hodgestar:
I'm not sure summit's implementation should be viewed as canonical and it
does seem odd to list contacts for some speakers and not others if the data
is available.

There is the concept of "corresponding author" but not sure how applicable
it is here / now.

CarlFK:
I have no problem with switching to another attribute name, like full_conf_url 
it is better than having code like this:

```                    # event['conf_url'] = 'https://summit.debconf.org' + row.find('conf_url').text
                    event['conf_url'] =   'https://za.pycon.org' + row.find('conf_url').text
```

hodgestar:
@carl -- shouldn't your code be something like
`urllib.basejoin(pentabarf_url, row.find('conf_url').text)`? That would
work for relative and absolute URLs and for both summit and pyconza.

drnlm:
I've add a 'full_conf_url' tag to the pentabarf.xml file which has the complete url now.

I've also added the twitter url for those persons that have that set as an attribute visible to authorized users.

hodgestar:
This will be complete when #122 lands.

hodgestar:
#122 has landed.

PR #121, opened 2015-09-24
Issue 119 more pentabarf

drnlm:
This should address issue #119

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
+++ b/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
@@ -55,7 +55,7 @@
                        <persons>
                            {% for author in items.item.talk.authors.all %}
                            {# person id is the author pk, which should be the right thing #}
-                               {% if user.is_staff and author == items.item.talk.corresponding_author %}
+                               {% if user.is_staff %}
                                {# We will want finer grained control off this eventually, but staff will do for now #}
                                <person contact="{{ author.email }}" id="{{ author.pk }}">{{ author.get_full_name|default:author.username }}</person>
                                {% else %}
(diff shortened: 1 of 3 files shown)


wafer/schedule/templates/wafer.schedule/penta_schedule.xml
```diff
@@ -70,7 +78,8 @@
                         <description/>
                         <type/>
                     {% endif %}
-                    <conf_url>{{ items.item.get_url }}</conf_url>
+                    {# forcing https here is a bit horrible - make this configurable somewhere? #}
+                    <conf_url>https://{{ WAFER_CONFERENCE_DOMAIN }}{{ items.item.get_url }}</conf_url>
```
stefanor: `{{ request.build_absolute_uri items.item.get_url }}` ?

drnlm: I'm pretty sure that won't do the right thing for the static site generation script, though.
We probably do want to add the pentabarf file to the static mirror eventually.

hodgestar: I'm okay with leaving the conference URL relative if that helps (since those loading the pentabarf presumably know the URL its relative to).

stefanor: > I'm pretty sure that won't do the right thing for the static site generation script, though.

Ah. That might be something to fix.

I'm OK with using WAFER_CONFERENCE_DOMAIN like this, we do in the e-mail templates, too.

hodgestar:
:+1:

drnlm:
Could use a sanity check before merging.

hodgestar:
Is `full_conf_url` something that exists in other pentabarf exports or something we're making up for ourselves? If we're making it up for ourselves, should we perhaps leave a comment to that effect in the template?

drnlm:
full_conf_url is something we're making up. The pentabarf format is undocumented.

We're not the first to do this. conf_url, as far as I can see, is something only summit produces - the original pentabarf only includes a 'slug' field for specifying conference details.

I'll add a comment, though

hodgestar:
:+1:

PR #122, opened 2015-09-26
released=True

CarlFK:


hodgestar:
What does `released=True` do? What consumes it and do any other pentabarf exports produce it?

CarlFK:
On Fri, Sep 25, 2015 at 9:19 PM, Simon Cross <notifications@github.com>
wrote:

> What does released=True do?
>
the upload script will not upload if False
It lets us hold back videos if someone asks.

assuming it gets hooked to a field first.

What consumes it
>
event['released'] = row.find('released').text == "True"


> and do any other pentabarf exports produce it?
https://summit.debconf.org/debconf15.xml
<released>True</released>



-- 
Carl K

drnlm:
:+1:

Issue #124, opened 2015-09-26
Switch to containerized builds.


hodgestar:

hodgestar:
Addressed by #125.

PR #125, opened 2015-09-26
Use containerized builds.

hodgestar:
Addresses #124. See http://docs.travis-ci.com/user/migrating-from-legacy/ for Travis documentation on migrating away from legacy builds.


drnlm:
:+1:

Issue #129, opened 2015-09-30
Pentabarf export calculates duration incorrectly for items spanning multiple slots


drnlm:
The pentabarf,xml export does the wrong thing if an item spans multiple slots. It calculates the duration from just the first slot, rather than the entire item.

drnlm:
#130 fixes this in the easy case.

Leaving this open for now so we can consider what to do if the slots aren't contigious.

drnlm:
The full schedule does the wrong thing with non-contigious slots already, and I'm not sure there's a good use case allowing a schedule item with non-contigious slots, so the obvious fix is just to make schedule validation in that case.

hodgestar:
I think there are some good use-cases for non-contiguous slots -- e.g. a long session than has a break for lunch, so it would be sad to lose that. There are various work-arounds though and this is a corner case, so perhaps there is something to be said for limiting options.

drnlm:
Since we currently don't properly support non-contigious slots, I think it's much better to fail sensibly rather than pretend to work and render an incorrect schedule. There also isn't going to be a good way of representing such an schedule item in the pentabarf export without creating virtual extra schedule items, since pentabarf assumes each item is a contigious block of time.

I really don't feel the complexity for supporting non-contigious slots correctly is worth while.

cateee:
In general, I think the concept of *slots* is wrong. A scheduled talk has a start date+time and and end date+time. The duration, and also the start/end time should be chosen freely by schedulers (coordinating with speaker), without need to align them in slots.  We already have 3 standard durations (45min [was the normal length], 20min ["short talks", possible the new default, 5min [lighting talks/live demos]), and I think wafer should handle the 3 of them easily (for submitter and scheduler): so without the need to generate new slots (specific for a room) and such not useful (IMO) overrides and technicalities. Additionally breakfast, meals, evening events and so on are handled differently, as exception in code, just because inflexibility of having slots.

About non-contiguous (I think this is the right spelling): I think it is a lot better to have multiple talks (as part I, part II, part III, ...) which allow more freedom to scheduler (different days) and submitters (having different description ("introduction/theory" vs. "real examples/problems").  Also the part of volunteers and video need to be aware of non-contiguous slots (volunteers could change between parts, video is recorded in different streams, ...). But also the grid schedule view would need a "(cont.)". So IMHO it would be much more complex to implement them, with very few advantages.

hodgestar:
[...]

PR #130, opened 2015-09-30
Bug/issue 129 fix scheduleitem duration

drnlm:
Partially fixes issue #129

If we have schedule items with slots that aren't contigious, this will do the wrong thing, but we already do very much the wrong thing in that case, so this is arguably not any worse.


hodgestar:
:+1:

Issue #132, opened 2015-10-03
Events in the schedule need to have people associated with them


drnlm:
Some events, such as the lightning talks, while not talks, still need speakers / session chairs and such in the schedule and pentabarf.xml export.

drnlm:
Closed by #133

PR #133, opened 2015-10-03
Feature/issue 132 add people to pages

drnlm:
Should solve #132

### DIFF
--- a/wafer/pages/models.py
+++ b/wafer/pages/models.py
@@ -68,6 +68,21 @@ class Page(models.Model):
             return True
         return False
 
+    def get_people_display_names(self):
+        names = []
+        for person in self.people.all():
+            full_name = person.get_full_name()
+            if full_name:
+                names.append(full_name)
+            else:
+                names.append(person.username)
+        if len(names) > 2:
+            # At some future point, I'll do something better
+            comma_names = ', '.join(names[:-1])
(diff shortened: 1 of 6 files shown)


wafer/schedule/templates/wafer.schedule/schedule_item.html
```diff
@@ -4,6 +4,9 @@
   by {{ item.talk.get_author_display_name }}
 {% elif item.get_url %}
   <a href="{{ item.get_url }}">{{ item.get_details|escape }}</a>
+  {% if item.page and item.page.people.count %}
```
hodgestar: We might want `.exists` here since that's probably cheaper for the db.

wafer/pages/models.py
```diff
@@ -62,9 +68,26 @@ def get_in_schedule(self):
...
+    def get_people_display_names(self):
+        names = []
+        for person in self.people.all():
+            full_name = person.get_full_name()
+            if full_name:
+                names.append(full_name)
+            else:
+                names.append(person.username)
+        if len(names) > 2:
+            # At some future point, I'll do something better
```
hodgestar: Maybe we should just remove this comment? I'm not sure what we can do that is better.

hodgestar:
Two minor comments. Otherwise looks good.

hodgestar:
:+1:

PR #137, opened 2015-10-03
Expose page details as description for pages in the schedule that have speakers, as being 'important' events

drnlm:
Mainly for the video use case


hodgestar:
:+1:

Issue #113, opened 2015-08-15
Wafer fails with latest django-registration-redux


drnlm:
From the docs of django-registration-redux 1.2:

"The templates in django-registration-redux assume you have a base.html template in your project’s template directory. This base template should include a title block and a content block. Other than that, every template needed is included. "

wafer's current base.html template lives in templates/wafer, which is not where django-registration-redux is looking for it. It also doesn't have a title block

drnlm:
We can probably address this by providing a custom registration_base.html template, which provides the appropriate title and content blocks

stefanor:
I cherry-picked 89b54ef274a3d0f4ddbedb18a3671043665b4235 into master, fixing this.

PR #140, opened 2015-10-17
Initial drag and drop schedule editor

zoidbergwill:
I've barely looked at this since the sprints, I've spotted a couple things I need to still do.

- [ ] Add link to edit schedule from schedule view in the site and admin control panel
- [x] Get rid of `print`s and `console.log`s
- [x] Make day configurable in editor
- [ ] Fix tests and maybe move some logic around.
- [ ] Add tests for custom API logic


zoidbergwill:
It looks like django-rest-framework doesn't support Django 1.6, which is EOL anyway.

wafer/schedule/views.py
```diff
@@ -203,3 +210,70 @@ def get_context_data(self, **kwargs):
...
+    """
+    queryset = ScheduleItem.objects.all()
+    serializer_class = ScheduleItemSerializer
+
+
+class ScheduleEditView(TemplateView):
+    template_name = 'wafer.schedule/edit_schedule.html'
+    date_field = 'date'
+    allow_future = True
+
```
drnlm: This view should be protected by appropriate admin permissions.

While the permissions on the rest API will prevent unauthorized changes, we don't want to leak unpublished information in the schedule carelessly.

wafer/schedule/views.py
```diff
@@ -203,3 +210,70 @@ def get_context_data(self, **kwargs):
...
+                            talk = schedule_item.talk
+                            aggregated_venue['title'] = talk.title
+                            aggregated_venue['talk'] = talk
+                            aggregated_venue['scheduleitem_id'] = talk.talk_id
+                        if schedule_item.page:
+                            page = schedule_item.page
+                            aggregated_venue['title'] = page.name
+                            aggregated_venue['page'] = page
+                            aggregated_venue['scheduleitem_id'] = page.id
+
```
drnlm: We should fileter out container pages - those with 'exclude_from_static' set to True - from this list.

wafer/schedule/views.py
```diff
@@ -203,3 +210,70 @@ def get_context_data(self, **kwargs):
...
+        day = days.first()
+        if id:
+            day = days.get(id=id)
+
+        accepted_talks = Talk.objects.filter(status=ACCEPTED)
+        venues = Venue.objects.filter(days__in=[day])
+        slots = Slot.objects.filter(day=day).select_related(
+            'day', 'previous_slot').prefetch_related(
+            'scheduleitem_set', 'slot_set')
```
drnlm: day may not be set for slots that have prev_slot set, so this can miss slots of interest.

This arguably should be fixed by changing how the slot model behaves, though.

zoidbergwill: Yeah, this isn't working for day 2, and I think it might be because of that. I'll look into it more.

wafer/settings.py
```diff
@@ -148,6 +148,8 @@
     'wafer.sponsors',
     'wafer.pages',
     'wafer.tickets',
+    # Django isn't finding the overridden templates
+    'registration',
```
hodgestar: Could you elaborate a bit on the moving of `registration` to after the `wafer.*` entries?

hodgestar:
Given that this is a new feature, I'm in favour of opening issues for the improvements we need and then landing this branch and working on things in smaller chunks.

If others are happy with that, I'm happy to create tickets and land this.

zoidbergwill:
Sounds good to me. I think I should just sort out the issue with not using [prev_slot](https://github.com/CTPUG/wafer/pull/140#discussion_r42491732), before we land it. I'll try get it done in the next few days.

wafer/settings.py
zoidbergwill: This might've just been an issue with the version of Django I was using, but there was an issue where Django was only finding the `registration` library's templates, and not our custom versions it should use.

drnlm: From the description, I suspect this is related to #113 

I haven't looked at the django-registration code to see why this fix works, and it may not be something we want to rely on, but this change doesn't seem particularly harmful.

drnlm:
:+1: on landing this early and fixing issues as we find them

wafer/settings.py
drnlm: Having looked more closely at what's happening,, this looks sane, and, as it relies on Django's documentated handling of INSTALLED_APPs, it should be safe enough to land.

wafer/schedule/views.py
zoidbergwill: This fixed day 2. \o/

wafer/schedule/views.py
zoidbergwill: Resolved in https://github.com/CTPUG/wafer/commit/57230b99f9a6028f41700a22036e8bcdd09f2d7e

hodgestar:
Landing -- outstanding issues will be addressed separately.

PR #142, opened 2015-11-17
Debian SSO

stefanor:
This PR first pre-factors our GitHub SSO, to improve error handling, and make it easier to implement more SSO backends. Then adds Debian's client-certificate SSO.


wafer/registration/sso.py
```diff
@@ -0,0 +1,146 @@
...
+                    first_name += u' ' + person['mn']
+                last_name = person['sn']
+                name = (first_name, last_name)
+                break
+
+    user = sso(identifier=identifier, desired_username=username, name=name,
+               email=email)
+
+    if not user.is_active:
+        raise SSOError('Account disabled')
```
drnlm: Since this "if not user.is_active:' construct is repeated in both implementation, should it not be folded into the sso function? Are there circumstances where sso will be called where this is not an error condition?

wafer/registration/sso.py
```diff
@@ -0,0 +1,146 @@
...
+    name = ('Unknown User', email)
+    if not get_user_model().objects.filter(**identifier).exists():
+        r = requests.get('https://nm.debian.org/api/people',
+                         params={'uid': username},
+                         headers={'Api-Key': settings.WAFER_DEBIAN_NM_API_KEY})
+        if r.status_code != 200:
+            raise SSOError('Failed to query nm.debian.org')
+        if 'r' not in r.json():
+            raise SSOError('Failed to parse nm.debian.org respnose')
+        # The API performs substring queries, so we need to find the correct
```
drnlm: We probably should log these failures, as we do in the github_sso function

drnlm:
Some small comments, but otherwise looks good to me

wafer/registration/sso.py
stefanor: If you have your own custom User model, then this check may not be wanted. But all our SSO code assumes you don't. So, yes factoring that out makes sense.

stefanor:
:bell:

drnlm:
:+1:

PR #163, opened 2015-11-28
Feature/issue 129 check for contigious slots

### LINKED ISSUE #129: Pentabarf export calculates duration incorrectly for items spanning multiple slots

drnlm:
This adds a check that schedule items do not have non-contiguous slots.

Closes #129


wafer/schedule/admin.py
```diff
@@ -174,10 +195,12 @@ class ScheduleItemAdmin(admin.ModelAdmin):
     def changelist_view(self, request, extra_context=None):
         extra_context = extra_context or {}
         # Find issues in the schedule
+        all_items=None
```
hodgestar: PEP8 wants spaces around `=`.

wafer/schedule/templates/admin/scheduleitem_list.html
```diff
@@ -28,6 +28,14 @@
              {% endfor %}
           </ul>
           {% endif %}
+          {% if errors.non_contiguous %}
+          <h3>{% trans "Items in the schedule with Non-Contiguous slots" %}</h3>
```
hodgestar: Small `n` and `c`?

hodgestar:
Minor nitpicks, otherwise looks good. :)

hodgestar:
:+1:

PR #165, opened 2015-11-28
Bug/issue 162 drop south

### LINKED ISSUE #162: Remove south support
We no longer support Django 1.6, so there's no reason to keep the support for south migrations.

drnlm:
This strips out the remaining support for South

Closes #162


hodgestar:
:+1:

Issue #157, opened 2015-11-28
Add tests for Django REST Framework API.


hodgestar:

hodgestar:
Addressed by #167.

hodgestar:
Addressed by #167.

PR #167, opened 2015-11-28
Add rest api tests

hodgestar:
Addresses #157.


drnlm:
:+1:

As a future enhancement, we may want to add some tests that hit the API without super user privileges, but that isn't critical.

hodgestar:
Test added. Ready for re-review.

drnlm:
:+1:

PR #171, opened 2015-11-29
Add missing change_talk permission to Talk.Meta

madduck:
The "change_talk" permission is used in the can_edit function, and also
in the management script that creates some standard roles, but it's
actually never defined. Herewith fixed.


madduck:
Ouch. Sorry about this, I should have run the test myself.

`change_talk` is a builtin permission: https://docs.djangoproject.com/en/1.8/topics/auth/default/#default-permissions

PR #172, opened 2015-11-30
talks: Give corresponding_author permission like other authors

madduck:
The corresponding author of a talk needs not be in the list of authors,
and this is a good separation, allowing conference organisers to create
talks for invited speakers etc.

However, post-creation, unless a staff member, the corresponding author
won't have edit access to the talk, which is herewith fixed.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -110,10 +110,16 @@ class Talk(models.Model):
     pending = property(fget=lambda x: x.status == PENDING)
     reject = property(fget=lambda x: x.status == REJECTED)
 
+    def _is_among_authors(self, user):
+        if self.corresponding_author.username == user.username:
+            return True
+        # not chaining with logical-or to avoid evaluation of the queryset
+        return self.authors.filter(username=user.username).exists()
+
     def can_view(self, user):
         if user.has_perm('talks.view_all_talks'):
             return True
-        if self.authors.filter(username=user.username).exists():
+        if self._is_among_authors(user):
             return True
         if self.accepted:
             return True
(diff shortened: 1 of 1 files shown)


stefanor:
The `corresponding_author` field is set to the person who submits the talk. And is displayed in schedules as the only speaker.

To have that person not be in `authors` we must probably display the names of all authors in schedules.

madduck:
Or just the first of the authors… but right now, the `corresponding_author` is not automatically made an author. So I could create a talk with you as the only author, and then I'd show up in the schedule and you'd show up in the talk details.

(`corresponding_author` should have been called `submitter`, I wonder if that is a trivial change that a migration can easily handle…)

stefanor:
> Or just the first of the authors

We aren't tracking any ordering of them.

madduck:
@stefanor There won't be a solution that fits everything. I do like the `corresponding_author` as it is right now and it'll be useful in the DebConf context.

For the schedule, where space is tight, I would suggest to pick the author with the lowest ID in the `talks_talk_authors` relation as the primary author. That way, the result is only really undefined if a talk is created with several authors right from the start, whereas it'll be well-defined if authors get added later. Then I'd use an algorithm such as the following for the schedule display:

    schedule_author = authors[0]
    if len(authors) == 2:
        schedule_author += " and %s" % authors[1]
    elseif len(authors) > 2:
        schedule_author += ", et al."

then put the entire list of authors into the mouseover text, as well as the talk details page.

For cases where this doesn't yield the desired result, we could provide an override box that will be displayed in the schedule instead. This would allow us to have talks by e.g. "Neil McGovern, DPL", as well as talks in the schedule by "The Release Team", if that is desired on the individual level.

What do you think?

madduck:
(It's a shame that Github doesn't let us "fork an issue").

Let me create a new issue with my previous comment and then we resume discussion there.

hodgestar:
:+1:

PR #173, opened 2015-11-30
talks: Display corresponding author

madduck:
As the corresponding author isn't even necessarily included in the authors list, and it's not clear from the authors list who the corresponding author is, make this field visible (read-only) to people who can edit a talk anyway.

### DIFF
--- a/wafer/talks/forms.py
+++ b/wafer/talks/forms.py
@@ -4,7 +4,7 @@ from django.utils.translation import ugettext as _
 
 from crispy_forms.bootstrap import FormActions
 from crispy_forms.helper import FormHelper
-from crispy_forms.layout import Submit, HTML
+from crispy_forms.layout import Submit, HTML, Div, Field
 from markitup.widgets import MarkItUpWidget
 
 from wafer.talks.models import Talk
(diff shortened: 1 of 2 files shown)


wafer/talks/forms.py
```diff
@@ -25,6 +25,24 @@ def __init__(self, *args, **kwargs):
...
+        # to change the data if someone bypasses (removes) the HTML disabled
+        # tag. The following is basically display-only as the input field
+        # has no name.
+        self.helper.layout.insert(TalkForm.Meta.fields.index('authors'),
+                Div(
+                    HTML('<label class="control-label" for="id_corresponding_author">Corresponding author: </label>'),
+                    Div(
+                        HTML('<input id="id_corresponding_author" class="textinput textInput form-control" type="text" value="%s" readonly/>'
+                            % instance.corresponding_author.get_full_name()),
+                        **{'class':'controls'}
```
stefanor: Why not say `class='controls'`?

wafer/talks/forms.py
```diff
@@ -25,6 +25,24 @@ def __init__(self, *args, **kwargs):
                             _('Delete')))))
         else:
             self.helper.add_input(submit_button)
+        # insert before authors
+        # TODO: this is *YUK*, but if we add the field to Meta.fields below,
+        # then we can disable the widget, sure, but it'd still load all the 
+        # data, and we'd need to validate that the submitted form isn't trying
+        # to change the data if someone bypasses (removes) the HTML disabled
+        # tag. The following is basically display-only as the input field
+        # has no name.
```
stefanor: Yeah, this is pretty yuck. Maybe we have outgrown ModelForm, and should add a custom form with suitable permission checking.

wafer/talks/forms.py
madduck: Because `Div()` doesn't take `**kwargs` but expects a dict…

wafer/talks/forms.py
madduck: I've since come up with a better way shown in 8805473d5082aa656e7446eb1e7042cf1fa56080. I didn't do this at this stage as I tried to minimise changes, but if refactoring is an option, I do think we can stay with ModelForm.

wafer/talks/forms.py
stefanor: You aren't passing a dict. You're creating a dict, and then passing it as keyword arguments.

madduck: Hm. I really don't remember what rode me, but you're obviously right.

hodgestar:
It will be a lot easier to just add the corresponding author to the edit template. I'll make a pull request for that.

hodgestar:
I vote for closing this once #180 lands if there are no objections.

hodgestar:
Closing for now. Feel free to re-open if needed.

Issue #117, opened 2015-09-15
Add organizer note space for talk abstracts


drnlm:
The talks current provide notes which are editable by the speaker and the conference organisers, which is useful for tracking various details, but being editable by the speaker, it's not ideal for to use for notes by the talk organiser about the talk.

Talks should have an extra "organiser notes" field which isn't editable by the speaker.

There may actually be a case for 2 such fields - one visible to the speaker and one for private notes only visible by the  talk committee (when there's a committee) and any mentors.

madduck:
Adding the field is trivial and done in https://github.com/madduck/wafer/commit/c65a299be3529fd0ae76fc35f088e476dbcef0ed. The field isn't visible to the user, but it is exposed in the admin interface.

However, I am not yet far enough into the code that i know how to tie it down security-wise, and I have not looked at giving access to talk admins or mentors. I'd love to be shoved in the right direction here.

If I add 'private_notes' to 'list_editable' in wafer.talks.admin.TalkAdmin, I get an error:
    <class 'wafer.talks.admin.TalkAdmin'>: (admin.E122) The value of 'list_editable[1]' refers to 'private_notes', which is not an attribute of 'talks.Talk'.
even though the field is obviously an attribute of Talk and the migration has been applied.

madduck:
What did you intend for the second field, i.e. a notes space visisble to both, submitter and the editors? Is this some sort of "chat" where the two can go back and forth? Why not just use the existing notes field? I also wonder if that's needed (submitters won't get notified…) or whether email will be used anyway.

drnlm:
The current notes field is intended for the speaker to add information that isn't relevant to the abstract

"I want to connect to the speaker system to play some videos", "This talk will be simalar to the talk I gave at X, but I'm going to spend less time on Y" and that sort of thing.

The organiser notes is intended for a reviewer to add notes about the submission , for the benefit of other reviewers - "I think the proposal is unclear, so I've asked for more information about X", "This talk is very similar to Z's proposal, so we should only accept one of them". While this can be handled by email, it is useful to keep relevant infromation together. This should generally not be exposed to the talk author, since it may leak information. 

There is a use case for a notes field that is only editable by the organisers and visible to the author for review comments. These should not be editable by the talk submitter for auditability, but this is only really useful if the talk review work-flow is largely web-driven.

madduck:
Thanks for the explanations, @drnml.

[...]

PR #175, opened 2015-12-01
Private talk notes

madduck:
This addresses #117 in what I hope is an acceptable way. I did have to refactor the crispy form code a bit, but I only really made it more explicit.

The trick to remove a field from `self.fields` to hide it *and* prevent access to it (incl. preventing it from being considered after submission) seems to be standard…

I chose to allow HTML in the private note space, and to format URLs for convenience. While this isn't 100% safe I am sure, we are giving access to orga folks here…

### DIFF
--- a/wafer/talks/forms.py
+++ b/wafer/talks/forms.py
@@ -12,7 +12,12 @@ from wafer.talks.models import Talk
 
 class TalkForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop('user')
         super(TalkForm, self).__init__(*args, **kwargs)
+
+        if not self.user.has_perm('talks.edit_private_notes'):
+            self.fields.pop('private_notes')
+
         self.helper = FormHelper(self)
         submit_button = Submit('submit', _('Submit'))
         instance = kwargs['instance']
@@ -28,7 +33,8 @@ class TalkForm(forms.ModelForm):
 
     class Meta:
         model = Talk
-        fields = ('title', 'talk_type', 'abstract', 'authors', 'notes')
+        fields = ('title', 'talk_type', 'abstract', 'authors', 'notes',
+                  'private_notes')
         widgets = {
             'abstract': MarkItUpWidget(),
             'notes': forms.Textarea(attrs={'class': 'input-xxlarge'}),
--- a/wafer/talks/templates/wafer.talks/talk.html
+++ b/wafer/talks/templates/wafer.talks/talk.html
@@ -41,6 +41,13 @@
 <div class="well">
     {{ object.abstract.rendered|safe }}
 </div>
+{% if perms.edit_private_notes and object.private_notes %}
+<h2>Private notes</h2>
+<p>(The following is not visible to submitters or attendees.)</p>
+<div class="well">
+  {{ object.private_notes|urlize|linebreaks }}
+</div>
+{% endif %}
 {% if talk.talkurl_set.all %}
 <div class="well" id="urls">
    <div>{% trans "URLS" %}</div>
--- a/wafer/talks/views.py
+++ b/wafer/talks/views.py
@@ -66,6 +66,11 @@ class TalkCreate(LoginRequiredMixin, CreateView):
     form_class = TalkForm
     template_name = 'wafer.talks/talk_form.html'
 
+    def get_form_kwargs(self):
+        kwargs = super(TalkCreate, self).get_form_kwargs()
+        kwargs['user'] = self.request.user
+        return kwargs
+
     def get_context_data(self, **kwargs):
         context = super(TalkCreate, self).get_context_data(**kwargs)
         context['can_submit'] = getattr(settings, 'WAFER_TALKS_OPEN', True)
@@ -88,6 +93,11 @@ class TalkUpdate(EditOwnTalksMixin, UpdateView):
     model = Talk
     form_class = TalkForm
     template_name = 'wafer.talks/talk_form.html'
+
+    def get_form_kwargs(self):
+        kwargs = super(TalkUpdate, self).get_form_kwargs()
+        kwargs['user'] = self.request.user
+        return kwargs
 
     def get_context_data(self, **kwargs):
         context = super(TalkUpdate, self).get_context_data(**kwargs)
--- a/wafer/talks/migrations/0003_talk_private_notes.py
+++ b/wafer/talks/migrations/0003_talk_private_notes.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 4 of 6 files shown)


wafer/talks/views.py
```diff
@@ -51,6 +52,7 @@ def get_object(self, *args, **kwargs):
         '''Only talk owners can see talks, unless they've been accepted'''
         object_ = super(TalkView, self).get_object(*args, **kwargs)
         if object_.can_view(self.request.user):
+            object_.private_notes = mark_safe(object_.private_notes)
```
stefanor: Why do we consider this safe?

wafer/talks/templates/wafer.talks/talk.html
```diff
@@ -41,6 +41,13 @@
 <div class="well">
     {{ object.abstract.rendered|safe }}
 </div>
+{% if perms.edit_private_notes and object.private_notes %}
+<h2>Private notes</h2>
+<p>(The following is not visible to submitters or attendees.)</p>
+<div class="well">
+  {{ object.private_notes|urlize|linebreaks }}
```
stefanor: We should presumably display the notes from the speaker to the conference managers, too.

wafer/talks/views.py
```diff
@@ -66,6 +68,11 @@ class TalkCreate(LoginRequiredMixin, CreateView):
     form_class = TalkForm
     template_name = 'wafer.talks/talk_form.html'
 
+    def get_form_kwargs(self):
+        kwargs = super(TalkCreate, self).get_form_kwargs()
+        kwargs.update({'user': self.request.user})
```
stefanor: There's simpler syntax: `kwargs['user'] = self.request.user`

wafer/talks/views.py
madduck: Between Django's "safe" concept, the fact that private notes are limited to organisers and the benefit of HTML in the notes (not at all required), I considered the attack vectors to be negligible. But we certainly don't have to allow HTML parsing.

stefanor: If we make it a markdown field, then it can have the same HTML permitted in the talk abstract. Presumably that's sufficient?

madduck: I am fine with markdown or plain text, too. Let's go plain text then…

wafer/talks/forms.py
```diff
@@ -12,26 +12,37 @@
...
     def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop('user')
         super(TalkForm, self).__init__(*args, **kwargs)
         self.helper = FormHelper(self)
-        submit_button = Submit('submit', _('Submit'))
+
+        actions = (Submit('submit', _('Submit')),)
+
+        if not self.user.has_perm('talks.edit_private_notes'):
+            self.fields.pop('private_notes')
```
hodgestar: I think if you pop the field first, you won't need to entirely replace the layout, e.g.
```
if not self.user.has_perm('talks.edit_private_notes'):
    self.fields.pop('private_notes')
self.helper = FormHelper(self)
actions = (Submit('submit', _('Submit')),)
... possibly add delete action ...
self.helper.layout.append(FormActions(*actions))
```

wafer/talks/forms.py
```diff
@@ -12,26 +12,37 @@
...
+        self.helper.layout = Layout(
+                field_layout,
+                FormActions(*actions)
+                )
 
     class Meta:
         model = Talk
-        fields = ('title', 'talk_type', 'abstract', 'authors', 'notes')
+        fields = ('title', 'talk_type', 'abstract', 'authors', 'notes',
+                'private_notes')
```
hodgestar: Do we need to indent this two more spaces?

wafer/talks/templates/wafer.talks/talk.html
hodgestar: I'm tempted to say we shouldn't display any notes here so that the individual talk pages closely resemble what is seen by ordinary users of the site (but happy for it to be included if the consensus is to display it).

wafer/talks/views.py
hodgestar: Plus one on plain text.

hodgestar:
@madduck: Thanks for working on this! I've made a few comments but generally looks good. If you're happy to update the PR when you get a chance, that would be cool. Otherwise I can make the tweaks.

wafer/talks/templates/wafer.talks/talk.html
madduck: I now made it configurable via `settings.py` but left the default at `True` (i.e. show private notes). I am not wed to this default setting.

wafer/talks/forms.py
madduck: Okay, even though I actually thought I made it a bit more readable. In the interest of small patches, I've tried your suggestion, it works, so it'll be pushed.

madduck:
@hodgestar I hope I addressed all the issues. A new branch was pushed and it's ready again for merging.

wafer/talks/templates/wafer.talks/talk.html
```diff
@@ -41,6 +41,13 @@
 <div class="well">
     {{ object.abstract.rendered|safe }}
 </div>
+{% if show_private_notes and perms.edit_private_notes and object.private_notes %}
```
hodgestar: Couple of thoughts on this:
* I'm still not wild about showing the private notes here. Anyone want to tie-break for madduck and I?
* I think the `<h2>` heading needs to be inside the `<div class="well">` for consistency with what happens with other headings generated by the talk abstract itself?

hodgestar:
Left last comment. Looks good otherwise though.

wafer/talks/templates/wafer.talks/talk.html
drnlm: I like having this information here so all everything related to the talk is immediately accessible for reviewers. If we're not going to show the notes here, we will need to document how people reviewing and commenting on the talks get access to this information - whether via the admin interface or via some other form.

hodgestar: @madduck Given @drnlm's vote, let's remove the setting and `show_private_notes` and go back to always showing it if the user has the `edit_private_notes` permission?

Apologies for the back and forth on this.

madduck: No worries, @hodgestar. Removing the setting is trivial, but is there a reason why not simply keep it and default it to `True`? Or do you just prefer the code to stay as lean as possible?

hodgestar: Yep -- just a desire to keep the code base lean where we can.

madduck:
Removed the `show_private_notes` setting from the branch. Hoping it's now ready…

hodgestar:
Looks good! Thanks!

hodgestar:
:+1:

Issue #128, opened 2015-09-29
Schedule should be more robust against slot deletion


drnlm:
Currently the schedule admin view assumes that every schedule item will have a slot. This isn't fully enforced, so actions like deleting a slot without editing the schedule items first can break this assumption.
This leads to bad stuff happening when then trying to manage the schedule items again.

We should either prevent deleting slots that are referred to by schedule items, or make schedule items more robust against this case. Arguably the second option is the better one.

drnlm:
#131 fixes the most obvious issues.

Leaving this open as we will also need to fix Venue and Day deletion. Those we should probably prevent if they have slots or schedule items assigned to them.

drnlm:
Everything should be addressed now.

PR #176, opened 2015-12-21
Protect scheduleitems when deleting days and venues

drnlm:
This addresses the remaining part of #128 

Days and Venues should not be deleted while scheduleitems refer to them, so we use Django's on_delete setting to ensure that doesn't happen.


hodgestar:
:+1:

PR #178, opened 2015-12-27
Display all a talk's authors

stefanor:


wafer/talks/models.py
```diff
@@ -74,20 +74,25 @@ def get_author_contact(self):
...
-        return self.corresponding_author.username
+    get_corresponding_author_name.admin_order_field = 'corresponding_author'
+    get_corresponding_author_name.short_description = 'Corresponding Author'
+
+    def get_authors_display_name(self):
+        authors = list(self.authors.all())
+        # Corresponding authors first
+        authors.sort(
+            key=lambda author: 0 if author == self.corresponding_author
+                               else author.userprofile.display_name())
```
hodgestar: I suspect using a mix of `0` and strings as a keys will fail on Python 3?

hodgestar:
Left one comment, otherwise looks good to me.

hodgestar:
:+1: if Travis builds pass.

PR #184, opened 2016-01-07
Load Twitter boilerplate only when username defined

madduck:
Unless the UserProfile specifies a Twitter username, there is no point
in loading the Twitter boilerplate. In fact, it could even be considered
wrong to do so… This makes loading the JavaScript conditional, just like
display of the Twitter buttons is conditional.


hodgestar:
Looks good to me.

:+1:

Issue #147, opened 2015-11-27
Key-value store for custom properties


madduck:
Over the years at DebConf, we kept adding fields and growing the database. Recently, the idea came up to offer a flexible key-value store instead, such that the database itself does not need to be touched and more can be done in the middleware or in scripts.

Think of this as "user tags" and it could be anything from the admins creating keys for people who need them, to every attendee having the ability to associate any data with any object (page, attendee, talk, etc.).

A first braindump of the idea is here:
https://wiki.debconf.org/wiki/Wafer/KeyValueStore

This is high-priority for DebConf, I'd say, and I am very motivated to work on it. Posting the issue here as an invitation for feedback and warnings etc.

stefanor:
@madduck: I see you have a branch. What's the status of it? And, does it have to use an incompatible license?

madduck:
The status of the branch is: theoretically ready, but the way I've done it means that `admin.py` integration won't be able to reuse code from `django.contrib.contenttypes` and I could not be bothered to hack that. Instead, I've written https://groups.google.com/forum/#!searchin/django-developers/krafft/django-developers/jx-14OddSqA/-ssU44MaCQAJ (and found https://groups.google.com/forum/#!searchin/django-users/genericforeignkey/django-users/4cMifVb7iDM/0_BS08RFtS0J) and was investigating trying to write `ExtendedGenericForeignKey`.

I don't see myself working on this until 2015-12-26 as we're now with family for the next few days. However, neither API nor database are expected to change as a result of this, so it should be possible to just work off what we have for now and we can later change the guts. I think.

The licence is just my default. It can be anything else. I don't see any reason why we should forego copyleft though. But I am not going to dig in my hoofs on this.

madduck:
I've reworked the code as suggested on django-developers, now using `IndirectGenericForeignKey`. This made the code more robust and more usable. It's still missing some tests and docs, but I'll get onto that (and the other TODO items) as soon as I know that this is the right track.

I've also updated the licence on all code to be explicitly the same licence as Wafer.

PR #186, opened 2016-01-13
147 kvpairs

madduck:
While there are still things left to do (cf. `docs/kvpairs.rst`), this is ready for merging. The licence is now the same as Wafer's. Please refer to #147 and the aforementioned documentation for more information. The code itself has also been documented.

Closes: #147

### DIFF
--- a/wafer/kvpairs/models.py
+++ b/wafer/kvpairs/models.py
@@ -1,0 +1,245 @@
+# -*- coding: utf-8 -*-
+# (c) 2015-16 martin f. krafft <madduck@debconf.org>
+# Released under the terms of the same licence as Wafer.
+
+from django.db import models
+from django.db.models import signals
+from django.core.exceptions import SuspiciousOperation, ValidationError
+from django.contrib.contenttypes.models import ContentType
+from django.utils.encoding import python_2_unicode_compatible
+
+from django.conf import settings
+KEYNAME_MAXLEN = settings.WAFER_KVPAIRS_KEYNAME_MAXLEN
+VALUE_MAXLEN = settings.WAFER_KVPAIRS_VALUE_MAXLEN
+AUTOCREATE_KEYS = settings.WAFER_KVPAIRS_AUTOCREATE_KEYS
+
+# the list of models to which keys can be "attached", not sure why we limit
+# it, but it feels right. This feeds the choices parameter of the Key.model_ct
+# field, and can be disabled right here…
+SUPPORTED_MODEL_TYPES = ('sites.site', 'auth.user', 'users.userprofile',
+        'pages.page', 'schedule.venue', 'schedule.scheduleitem', 'talks.talk',
+        'sponsors.sponsorshippackages', 'sponsors.sponsor',
+        'tickets.tickettype', 'tickets.ticket')
+
+
+@python_2_unicode_compatible
+class Key(models.Model):
+    """Definition of a specific key related to another model"""
+
+    # Key ownership, which controls who can access/modify the values
+    owner = models.ForeignKey(settings.AUTH_USER_MODEL,
+            on_delete=models.PROTECT, # could be SET(get_sentinel_user)…
+            verbose_name='Key owner')
+    group = models.ForeignKey('auth.Group', null=True, # groups are not mandatory
+            on_delete=models.SET_NULL,
+            verbose_name='Key owner group')
+
+    # The contenttypes framework already has all the models, let's link to
+    # that:
+    def _get_supported_models_as_q():
+        label_model_pairs = [s.split('.') for s in SUPPORTED_MODEL_TYPES]
+        if not label_model_pairs:
+            return models.Q()
+
+        acc = None
+        for a,m in label_model_pairs:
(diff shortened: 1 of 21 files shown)


madduck:
I've spent a bit of time researching the failing tests.

Regarding the failed Python2.7/Django1.7 test, all I can say is that I've developed this for Django 1.8, and if Dango 1.7 compatibility is a requirement (1.9 is out!), then I'll have to invest a couple of hours to work around some of the limitations the old code brings about.

Regarding the failing Python3.x tests, I am a bit at a loss, but I am hoping these are simple syntax issues.

drnlm:
While Django 1.7 support is somewhat debatable, the other python 3 test failures do need to be addressed. This feature does not justify any regressions in our support for python 3.

madduck:
@drnlm I completely agree that the Py3 issues need fixing, I just don't have any experience there yet, and I hope it's a trivial thing.

What do you suggest regarding Django 1.7? It's a sizeable amount of work required, I think. On the other hand, it'll likely lead to more robust code. I just don't know if it makes sense to block on it now…

drnlm:
Python 3 dropped python 2's implicit relative import logic. You'll need to use something like
'from wafer.kvpairs.fields import x' everywhere to solve the python 3 import issues.

There are a number of other python 3 issues with this branch, though - print function, various changed imports (reduce, for example) and so forth.

madduck:
@drnlm The Python3 problems are all solved. I am hoping to address Django 1.7 and 1.9 issues with a simpler approach.

madduck:
Just FTR, I am not making the desired progress with the simpler approach. I can make it work for Django 1.7 and 1.8, or for Django 1.9 and hack it for 1.8. I don't want to write spaghetti code in a vague attempt to support all three versions.

Given the recent Dj1.9 compatibility commits, couldn't we just target Wafer 0.4 at Django 1.9 and be done with it?

drnlm:
Dropping support for Djanog 1.7 is fine. Django 1.8 is a long term support release, so I'm very reluctant to drop compatibility with it so early on.

wafer/kvpairs/models.py
```diff
@@ -0,0 +1,252 @@
...
+@python_2_unicode_compatible
+class Key(models.Model):
+    """Definition of a specific key related to another model"""
+
+    # Key ownership, which controls who can access/modify the values
+    owner = models.ForeignKey(settings.AUTH_USER_MODEL,
+            on_delete=models.PROTECT, # could be SET(get_sentinel_user)…
+            verbose_name='Key owner')
+    group = models.ForeignKey('auth.Group', null=True, # groups are not mandatory
+            on_delete=models.SET_NULL,
```
drnlm: If group is optional, it also have blank=True, otherwise form validation will fail if it's unset, no?

wafer/kvpairs/models.py
```diff
@@ -0,0 +1,252 @@
...
+                self.key.name, self.value)
+
+    def get_model_class(self):
+        '''Returns the model class associated with the Key'''
+        return self.key.get_model_class()
+
+    def get_ref_instance(self):
+        '''Returns the instance referenced by the KeyValuePair'''
+        model_class = self.get_model_class()
+        return model_class.objects.get(pk=self.ref_id)
```
drnlm: This looks a lookup that could be cached - we don't expect the referenced object to change and this seems like it will be called reasonably often if working with the value pairs.

drnlm:
So this works as expected, but the admin interface is horribly clunky.

What I'm not clear on is quite how this is meant to be used. For the documentation, it doesn't look like the intention is to manage these through the admin interface, but it's not clear quite what the approach will be. Will each app using these have to write forms to handle the keys and key values? How will these look? Perhaps you could clarify this some more?

wafer/kvpairs/models.py
madduck: I always try to avoid storing blank values that are not NULL. Form validation seems not to trip over this, at least not in the admin interface…

wafer/kvpairs/models.py
madduck: Sounds good. I'll have to read up on how to do this though…

madduck:
@drnlm My plan was to enable inlines in the admin interface, so that you can add kvpairs for e.g. Talks right there on the talk's page. I am not there yet, obviously, but deriving from `GenericForeignKey` should make this easier.

wafer/kvpairs/models.py
drnlm: a) blank has nothing to do with database storage - it's purely a hint about validation

b) By default, the required attributed for model field is set from blank (``'{required': not self.blank}`` in django/db/models/fields/) . You have blank=False and you're not overriding the admin form to manipulate the required property, so how is the admin form allowing empty value?

madduck: You know this better than I do, so thanks for your feedback. Sounds like `blank` is needd. The admin form gets a select box for the group, and does not offer "(blank)", and so I was misled into thinking that it didn't accept blank values, but form validation is of course separate…

hodgestar:
Should we close this given that we landed #198?

PR #187, opened 2016-01-14
Feature/minimal django 1.9

drnlm:
This enables testing against Django 1.9 and adds a minimal set of fixes so the test suite passes.

I haven't tested anything further than running the test suite, though, so this probably isn't everything needed to support Django 1.9


hodgestar:
:+1:

PR #188, opened 2016-01-15
Disable NULL on n:m fields (pages, sponsors)

madduck:
This patch removes the null=True attributes from the ManyToManyFields
found in the pages.Page, sponsors.Sponsor and sponsors.Sponsorpackage
models, and therefore silences the following Django warnings:

WARNINGS:
pages.Page.files: (fields.W340) null has no effect on ManyToManyField.
pages.Page.people: (fields.W340) null has no effect on ManyToManyField.
sponsors.Sponsor.files: (fields.W340) null has no effect on ManyToManyField.
sponsors.SponsorshipPackage.files: (fields.W340) null has no effect on ManyToManyField.

### DIFF
--- a/wafer/pages/migrations/0003_non-null_people+files.py
+++ b/wafer/pages/migrations/0003_non-null_people+files.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pages', '0002_page_people'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='page',
+            name='files',
+            field=models.ManyToManyField(help_text='Images and other files for use in the content markdown field.', related_name='pages', to='pages.File', blank=True),
+        ),
+        migrations.AlterField(
+            model_name='page',
+            name='people',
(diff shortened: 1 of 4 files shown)


madduck:
All tests pass on my local system when run against Django 1.8 with either Python 2.7, or 3.4; I don't know what Travis' problem is, sorry…

hodgestar:
I'm not 100% why Travis is giving the error it is either, but Travis tests do need to pass before this lands.

madduck:
I would not be surprised if merging #189 also took care of these issues, as they do seem to be rooted in the migration of Schedule…

drnlm:
The crash is because of https://code.djangoproject.com/ticket/24513 , which was fixed in Dajngo 1.8.1, which is why it doesn't show up in local testing.

The Django version restriction in .travis.yml is too tight, evaluting to 'pip install Django<-1.8', which installs Django-1.8.0, rather than any of the 1.8 bugfix releases.

madduck:
Do you want me to patch `travis.yml` as part of this pull request?

drnlm:
My personal reaction is that fixing .travis.yml should be it's own pull request. It's a completely separate bug in our testing setup to this work.

madduck:
Done with #195.

madduck:
Thanks to @drnlm, this is now ready to go…

drnlm:
:+1:

PR #189, opened 2016-01-15
Provide migrations for Schedule changes

madduck:
Commit 88e341ff changed the on_delete parameter of the Slot.day and
ScheduleItem.venue fields, which requires a migration herein provided.

Signed-off-by: martin f. krafft <madduck@madduck.net>

PR #190, opened 2016-01-15
Remove 'ordering' model option in favour of 'order_with_respect_to'

madduck:
Prompted by this error:

  schedule.Slot: (models.E021) 'ordering' and 'order_with_respect_to'
  cannot be used together.

I found

  https://docs.djangoproject.com/en/1.8/ref/models/options/#order-with-respect-to

where it says:

  order_with_respect_to implicitly sets the ordering option

  Internally, order_with_respect_to adds an additional field/database
  column named _order and sets the model’s ordering option to this
  field. Consequently, order_with_respect_to and ordering cannot be used
  together, and the ordering added by order_with_respect_to will apply
  whenever you obtain a list of objects of this model.

This was apparently introduced in 1.8, and so I removed 'ordering'.

I am not sure what this does wrt Django 1.7, but I don't think this is
worth extra effort.


hodgestar:
Closing in favour of #191.

madduck:
#191 does not fix this, @hodgestar. #191 is just needed once #189 and #190 have been applied.

drnlm:
We need a commit along these lines for Django 1.9, but I think the correct solution is to remove order_with_respect_to in favour of ordering, which I think describes the desired behaviour better.

PR #191, opened 2016-01-15
Merge migrations of #189 and #190

madduck:
These two pull requests create leaf nodes which are merged herein.


hodgestar:
Oh, I see that this doesn't actually replace #189 and #190. Merging like this is not a good idea though. There should either just be one migration or we should number the migrations `3` and `4` instead of `3` and `3`.

Do you want to do that and re-ping this for review when that's done?

madduck:
Right. So let's be pragmatic and either make #190 depend on #189 (linearising), or merge the two unrelated changes into one changeset. Which would you prefer?

(I deliberately kept them separate, but if you would prefer such pull requests to be linearised right away in the future, I can certainly do that)

hodgestar:
Happy with any of the linearising options. Perhaps one pull request is
easiest?

madduck:
@hodgestar, might I suggest that you merge #189 and then I push an updated #190 and we close #191 (herewith done)?

hodgestar:
Sounds good.

PR #193, opened 2016-01-24
Bugs/issue 192 unicode calls

### LINKED ISSUE #192: Remove'unicode' calls from wafer
Current wafer using python 3 fails on several admin tasks because ```UserProfile.__str__``` tries to call ```unicode```, which is obviously not defined.

We should handle the difference between python 2 and python 3 correctly in this situation.

There are a couple of other calls to unicode() that look dangerous in the error paths in /registration/views.py that should probably be fixed as well.

drnlm:
Ensure that we only call unicode when running on Python 2.

Should close #192

### DIFF
--- a/wafer/registration/views.py
+++ b/wafer/registration/views.py
@@ -1,4 +1,6 @@
 import urllib
+import sys
+PY2 = sys.version_info[0] == 2
 
 from django.contrib.auth import login
 from django.contrib import messages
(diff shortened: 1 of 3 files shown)


madduck:
Isn't this what [the python_2_unicode_compatible decorator](https://docs.djangoproject.com/en/1.9/ref/utils/#django.utils.encoding.python_2_unicode_compatible) is for? See https://docs.djangoproject.com/en/1.9/topics/python3/#str-and-unicode-methods?

drnlm:
python_2_unicode_compatible makes it easier to support python 2 and python 3, but it is not magic. It does not define unicode on python 3, nor does it ensure that calling ``str()`` within a the ``__str__`` method handles the differences between python 2 and python 3. That logic still needs to be provided by us.

stefanor:
:+1:

PR #194, opened 2016-01-30
Feature/add django reversion support

drnlm:
This uses django-reversion to track edits to talks and pages.

The history is only exposed through the admin interface. We probably want to expose it in some way for people editing pages, but I'm not sure that's required for talks. Displaying the date of last edit to people who can change talks and pages is probably a good idea.

We may want to hook this up to the Sponsors model as well. I'm not sure it makes sense for the other models we have.

### DIFF
--- a/wafer/talks/views.py
+++ b/wafer/talks/views.py
@@ -78,19 +78,19 @@ class TalkCreate(LoginRequiredMixin, CreateView):
         context['can_submit'] = getattr(settings, 'WAFER_TALKS_OPEN', True)
         return context
 
+    @revisions.create_revision()
     def form_valid(self, form):
         if not getattr(settings, 'WAFER_TALKS_OPEN', True):
             raise ValidationError  # Should this be SuspiciousOperation?
         # Eaaargh we have to do the work of CreateView if we want to set values
         # before saving
-        with revisions.create_revision():
-            self.object = form.save(commit=False)
-            self.object.corresponding_author = self.request.user
-            self.object.save()
-            revisions.set_user(self.request.user)
-            revisions.set_comment("Talk Created")
-            # Save the author information as well (many-to-many fun)
-            form.save_m2m()
+        self.object = form.save(commit=False)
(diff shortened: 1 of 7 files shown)


wafer/talks/views.py
```diff
@@ -104,12 +109,30 @@ def get_context_data(self, **kwargs):
...
 
+    def form_valid(self, form):
+        # Is there a better way to do this?
+        # I don't want to use RevisionMiddleware, because I think we
+        # want more control over things than that provides, but
+        # this feels clunky
+        with revisions.create_revision():
+            result = super(TalkUpdate, self).form_valid(form)
+            revisions.set_user(self.request.user)
+            revisions.set_comment("Talk Modified")
```
hodgestar: I'm not sure how much better this is, but I think we could do:
```python
@revisions.create_revision()
def form_valid(self, form):
    revisions.set_user(self.request.user)
    revisions.set_comment("Talk Modified")
    return super(TalkUpdate, self).form_valid(form)
```

hodgestar:
Left a couple of points for discussion.

Agreed that we should add it for Sponsors (and Sponsor Packages) and that a separate PR is likely better for that (although I don't really mind if we do it all in one go either).

wafer/talks/views.py
drnlm: I'm not sure how I missed the decorator option in the reversion docs.

While this is a bit more compact, I'm mainly concerned about shoving everything into form_valid. Unfortunately, I don't see a better place for this, especially as I don't want to go the ReversionMiddlewar route.

drnlm:
Since we only currently manage Sponsors through the admin interface, it's easy enough to hook up as well, so I've done that.

I've changed to the decorator syntax, since it is a bit more readable and there's no benefit in our use case for using the context manager syntax.

After re-reading the Django docs, I haven't found an obviously better way of hooking things up, so I've dropped that comment.

hodgestar:
:+1:

PR #195, opened 2016-02-03
Loosen Django version requirement spec

madduck:
PR #188 exposed a Django version requirement in the travis spec that was
too tight. By converting from an inclusive ordered matching specifier to
the compatibility specifier, we ensure that e.g. Django 1.8.1 (or
whatever is current) will be used to fulfill a Django 1.8 dependency.

Ref: https://www.python.org/dev/peps/pep-0440/#compatible-release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
 # command to install dependencies,
 install:
    # Install the right version of Django first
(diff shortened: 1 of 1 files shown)


drnlm:
:+1:

PR #196, opened 2016-02-04
Feature/select2 support

### LINKED ISSUE #182: Add select2 widget for selecting talk authors and corresponding author.
See https://select2.github.io/examples.html#multiple for an example of how this might look.

We could use either

* https://www.djangopackages.com/packages/p/django_select2/ or
* https://www.djangopackages.com/packages/p/django-easy-select2/

not sure which is better.

drnlm:
Closes #182 using django-easy-select2. Django-select2 looks rather heavier than we require.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,7 @@ REQUIRES = [
     'requests',
     'django-medusa>=0.3.0',
     'django-reversion>=1.10',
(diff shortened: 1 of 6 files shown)


stefanor:
:+1:

PR #198, opened 2016-02-07
Add simple key-value store, and attach to users

stefanor:
Fixes: #147

Simple KeyValue store implementation. Currently the admin view of this is hacky-as-hell. You can't nest inlines (and these models don't really work well in inlines, anyway), so we're probably going to have to do some major customization of the user edit form, to handle these well.

### DIFF
--- a/wafer/users/admin.py
+++ b/wafer/users/admin.py
@@ -5,15 +5,48 @@ from django.contrib.auth import get_user_model
 from wafer.users.models import UserProfile
 
 
+# User-centric
+
 class UserProfileInline(admin.StackedInline):
     model = UserProfile
     can_delete = False
     verbose_name_plural = 'profile'
+    exclude = ('kv',)
 
 
 class UserAdmin(UserAdmin):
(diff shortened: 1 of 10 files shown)


wafer/users/admin.py
```diff
@@ -5,15 +5,48 @@
...
+    return obj.user.email
+
+
+class UserProfileAdmin(admin.ModelAdmin):
+    model = UserProfile
+    readonly_fields = ('user',)
+    exclude = ('kv',)
+    inlines = (KVPairsInline,)
+    list_select_related = ('user',)
+    list_display = (username, 'display_name', email)
```
hodgestar: Do `user__username` and `user__email` not work in `list_display`?

stefanor: Nope.

hodgestar:
:+1: from me, although I'm a bit unsure about why the last commit doesn't have a build.

stefanor:
Because GitHub orders by commit date, not graph order.

Issue #181, opened 2015-12-27
And Django Restforms API to the core models.


hodgestar:
We'd like to make it possible for other software to interact with Wafer's core modules. This issue is for adding basic Django Restforms API support.

Future PRs will need to address:

* Cleaning up permissions (mostly just checking that the permissions we have in place make sense in the context of access via the API)
* Supporting the new key-value store.

drnlm:
With #200 merged, this is largely done.

We should look at the integration of the kv-pairs work and the REST api before releasing 0.4, though.

drnlm:
Basic support for kv-pairs has been added with #235 , although there is scope for improvement.

The entire API still needs documentation.

I'd appreciate someone sanity checking the current permissions, especially around the kv stuff. We may also want to limit the kv's shown on other objects to those in the viewers group, as for the kv api, to prevent undue information leakage.

hodgestar:
#547 added an API for tickets and ticket types.

PR #200, opened 2016-02-10
Feature/rest api

drnlm:
This is a minimalist stab at adding REST support to the core models, as requested in issue #181 .

There are open questions around the permissions on different parts of the API, and on what we should expose via the API. As it stands, this probably exposing too much, and tighter permissions checking would be a good idea, especially on the user api.


madduck:
If I may offer 2¢: I am a bit scared by "tighter permissions checking would be a good idea". Wouldn't it be better to deny-all and selectively add access, rather than to attempt to tighten the security post-hoc?

wafer/pages/serializers.py
```diff
@@ -0,0 +1,24 @@
...
+    @revisions.create_revision()
+    def create(self, validated_data):
+        revisions.set_comment("Created via REST api")
+        return super(PageSerializer, self).create(validated_data)
+
+    @revisions.create_revision()
+    def update(self, page, validated_data):
+        revisions.set_comment("Changed via REST api")
+        page.parent = validated_data['parent']
+        page.content = validated_data['content']
```
hodgestar: Do we want to allow updating of other attributes like `include_in_menu`, `exclude_from_static` and `people`? Maybe `files` too but perhaps that should be left for a later pull request.

wafer/pages/views.py
```diff
@@ -49,3 +52,10 @@ def slug(request, url):
         return EditPage.as_view()(request, pk=page.id)
 
     return ShowPage.as_view()(request, pk=page.id)
+
+
+class PageViewSet(viewsets.ModelViewSet):
+    """API endpoint for users."""
+    queryset = Page.objects.all()
+    serializer_class = PageSerializer
+    permission_classes = (DjangoModelPermissions, )
```
hodgestar: Do we perhaps want to use `DjangoModelPermissionsOrAnonReadOnly` since anonymous users can read all pages from the site already?

wafer/sponsors/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+    @revisions.create_revision()
+    def create(self, validated_data):
+        revisions.set_comment("Created via REST api")
+        return super(SponsorSerializer, self).create(validated_data)
+
+    @revisions.create_revision()
+    def update(self, sponsor, validated_data):
+        revisions.set_comment("Changed via REST api")
+        sponsor.name = validated_data['name']
+        sponsor.description = validated_data['description']
```
hodgestar: Do we want to allow `number_available`, `currency`, `price` and `short_description` to be updated too?

wafer/sponsors/views.py
```diff
@@ -20,3 +23,15 @@ class SponsorView(DetailView):
 class ShowPackages(ListView):
     template_name = 'wafer.sponsors/packages.html'
     model = SponsorshipPackage
+
+
+class SponsorViewSet(viewsets.ModelViewSet):
+    """API endpoint for users."""
+    queryset = Sponsor.objects.all()
+    serializer_class = SponsorSerializer
```
hodgestar: Do we want to add `permission_classes = (DjangoModelPermissionsOrAnonReadOnly, )` here?

wafer/sponsors/views.py
```diff
@@ -20,3 +23,15 @@ class SponsorView(DetailView):
...
+
+
+class PackageViewSet(viewsets.ModelViewSet):
+    """API endpoint for users."""
+    queryset = SponsorshipPackage.objects.all()
+    serializer_class = PackageSerializer
```
hodgestar: And perhaps here?

wafer/pages/serializers.py
drnlm: People should be added. I'm not sure about the other 2 attributes, as they're not attributes that should change often, and ideally are only set on page creation. I can easily be talked into adding support for them - I'm just not sure it's required.

wafer/talks/views.py
```diff
@@ -141,3 +146,35 @@ def get_context_data(self, **kwargs):
...
+    queryset = Talk.objects.all()
+    serializer_class = TalkSerializer
+    permission_classes = (DjangoModelPermissions, )
+
+
+    def list(self, request):
+        # We override the default implementation to only show accepted talks
+        # to people who aren't part of the management group
+        if Talk.can_view_all(request.user):
+            queryset = Talk.objects.all()
```
hodgestar: It's probably safer to do this by defining `def get_queryset(self, request):`. To use the permissions class, I think then one has to keep `queryset = Talk.objects.none()` so that the permissions understand which model class is being used.

wafer/talks/views.py
```diff
@@ -141,3 +146,35 @@ def get_context_data(self, **kwargs):
             user__talks__status='A').distinct().prefetch_related('user')
         context["speaker_rows"] = self._by_row(speakers, 4)
         return context
+
+
+class TalksViewSet(viewsets.ModelViewSet):
+    """API endpoint that allows talks to be viewed or edited."""
+    queryset = Talk.objects.all()
+    serializer_class = TalkSerializer
+    permission_classes = (DjangoModelPermissions, )
```
hodgestar: Same suggestion to allow anonymous users to retrieve (accepted) talks here.

wafer/pages/views.py
drnlm: Probably. Internal details like "exclude_from_static" and "show_in_menu" are currently only visible via the admin interface, which is why I didn't use that, but I don't see any significant implications to making that information available.

wafer/users/views.py
```diff
@@ -53,3 +56,11 @@ class EditProfileView(EditOneselfMixin, UpdateView):
...
     def get_success_url(self):
         return reverse('wafer_user_profile', args=(self.object.user.username,))
+
+
+class UserViewSet(viewsets.ModelViewSet):
+    """API endpoint for users."""
+    queryset = get_user_model().objects.all()
+    serializer_class = UserSerializer
+    # We want some better permissions than the default here, but
+    # IsAdminUser will do for now.
```
hodgestar: We don't seem to actually set the permissions to `IsAdminUser` here? The docs seem to indicate that the permissions might default to `AllowAny` unless configured otherwise in settings? Either way, seems good to be explicit.

wafer/pages/serializers.py
hodgestar: Seems nice to be able to update them.

hodgestar:
Left some comments and suggestions, but looks pretty cool.

wafer/users/views.py
drnlm: It's inheriting the default permission from settings. It's probably not a bad idea to set it explicitly here (and for the Schedule REST stuff, where the same consideration applies).

hodgestar:
:+1: on landing this and addressing any further improvements in separate issues / pull-requests.

PR #201, opened 2016-02-15
Wafer has some inappropriate defaults

stefanor:
Thanks to its pycon.za heritage.


drnlm:
One comment, but otherwise looks good to me.

stefanor:
That's a good point. The empty admins list did have a purpose

PR #203, opened 2016-02-17
Feature/revision compare support

drnlm:
This adds basic support for comparing the current revision with past revisions, since that's a useful feature.

The canonical compare implementation for django-reversion is django-reversion-compare, but that's GPL licensed, so I've opted not to go down that path.

This isn't particularly featureful, and the resulting output view could use some UI love, but works reasonably well.

Currently it's hooked up for Talks and Pages, as useful test cases.

### DIFF
--- a/wafer/compare/admin.py
+++ b/wafer/compare/admin.py
@@ -104,10 +104,9 @@ class CompareVersionAdmin(VersionAdmin):
                 continue
             else:
                 patch = generate_patch_html(revision, current, field)
-            the_diff.append('<h2>%s</h2>\n%s' % (field, patch))
+            the_diff.append((field, patch))
 
-        if not the_diff:
-            the_diff = ['<p>No differences found</p>']
+        the_diff.sort()
 
         context = {
             "title": _("Comparing current %s with revision created %s") % (
@@ -116,7 +115,7 @@ class CompareVersionAdmin(VersionAdmin):
             "opts": opts,
             "compare_list_url": reverse("%s:%s_%s_comparelist" % (self.admin_site.name, opts.app_label, opts.model_name),
                                                                   args=(quote(object_id),)),
-            "diff": '\n<p>\n'.join(the_diff),
+            "diff_list": the_diff,
         }
 
         extra_context = extra_context or {}
--- a/wafer/compare/admin.py
+++ b/wafer/compare/admin.py
@@ -85,15 +85,21 @@ class CompareVersionAdmin(VersionAdmin):
             # between the creation of revisions. This isn't ideal,
             # but should not be a fatal error.
             # Log this?
+            missing_field = False
             try:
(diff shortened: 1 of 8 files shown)


wafer/compare/admin.py
```diff
@@ -0,0 +1,155 @@
+# hack'ish support for comparing a django reversion
+# hisoty object with the current state
```
hodgestar: s/hisoty/history/

wafer/settings.py
```diff
@@ -149,6 +149,7 @@
     'wafer.tickets',
     # Django isn't finding the overridden templates
     'registration',
+    'wafer.compare',
```
hodgestar: Should we put this immediately below the other wafer apps?

wafer/compare/admin.py
```diff
@@ -0,0 +1,155 @@
...
+        opts = self.model._meta
+        object_id = unquote(object_id)
+        # get_for_object's ordering means this is always the latest revision.
+        current = self.revision_manager.get_for_object_reference(self.model, object_id)[0]
+        # The reversion we want to compare to
+        revision = self.revision_manager.get_for_object_reference(self.model, object_id).filter(id=version_id)[0]
+        the_diff = []
+        dmp = diff_match_patch()
+
+        for field in current.field_dict:
```
hodgestar: Do we need to do something like `for field in set(current.field_dict.keys()) | set(revision.field_dict.keys())` to get all possible fields in cases where fields have been removed from the current version?

Do we want to order the fields so that they're displayed consistently?

wafer/settings.py
drnlm: Seems sensible. Changed

wafer/compare/admin.py
drnlm: Extending the field list with those from the old version is probably sensible.

All the diffs are sorted (line 115) before being passed to the template, so the ordering should always be consistent.

wafer/compare/admin.py
```diff
@@ -0,0 +1,155 @@
...
+            except KeyError:
+                cur_val = "No such field in latest version"
+                missing_field = True
+            try:
+                old_val = revision.field_dict[field] or ""
+            except KeyError:
+                old_val = "No such field in old version"
+                missing_field = True
+            if missing_field:
+                diffs = dmp.diff_main(old_val, cur_val)
```
hodgestar: I imagine this diff could get confusing in cases where one of the values is missing and the other is something that contains words in common with the `No such field in X version` string. Do we perhaps want to construct the diff manually?

hodgestar:
Left a few comments but otherwise looks good. Woot.

wafer/compare/admin.py
drnlm: Fair point. Changed.

hodgestar:
:+1:

PR #204, opened 2016-03-05
Registration forms

stefanor:
This adds registration forms, as a separate registration mechanism.

Data is saved into KeyValue pairs.

This particular example form should probably be genericised (wafer shouldn't care about DebConf specifics).

The KV API clearly needs work. Updating contents is painful :(

### DIFF
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -258,10 +258,12 @@ WAFER_HIDE_LOGIN = False
 WAFER_TALKS_OPEN = True
 
 # Set this to False to disable registration
-# or 'pre' to enable pre-registration only
 WAFER_REGISTRATION_OPEN = True
+# Can be 'ticket' for Quicket tickets or 'form' for a classic form
+WAFER_REGISTRATION_MODE = 'ticket'
 
-# pre-registration requires a form
+# For REGISTRATION_MODE == 'form': The form to present, and the Group to
+# associate the responses with in the KV store
 WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
 WAFER_REGISTRATION_GROUP = 'Registration'
 
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -258,6 +258,7 @@ WAFER_HIDE_LOGIN = False
 WAFER_TALKS_OPEN = True
 
 # Set this to False to disable registration
+# or 'pre' to enable pre-registration only
 WAFER_REGISTRATION_OPEN = True
 
 # pre-registration requires a form
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -262,10 +262,8 @@ WAFER_REGISTRATION_OPEN = True
 # Can be 'ticket' for Quicket tickets or 'form' for a classic form
 WAFER_REGISTRATION_MODE = 'ticket'
 
-# For REGISTRATION_MODE == 'form': The form to present, and the Group to
-# associate the responses with in the KV store
+# For REGISTRATION_MODE == 'form', the form to present
 WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
-WAFER_REGISTRATION_GROUP = 'Registration'
 
 # Ticket registration with Quicket
 # WAFER_TICKET_SECRET = "i'm a shared secret"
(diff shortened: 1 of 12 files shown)


wafer/tickets/templates/wafer.tickets/claim.html
```diff
@@ -2,9 +2,9 @@
 {% load i18n %}
 {% load crispy_forms_tags %}
 {% block content %}
-{% if not WAFER_REGISTRATION_OPEN %}
-  {% trans "Registration is not currently open" %}
-{% elif not can_register %}
+{% if WAFER_REGISTRATION_OPEN != True %}
```
drnlm: If WAFER_REGISTRATION_OPEN is no longer going to be a simple boolean flag, we should rename it to something more descriptive. Something like WAFER_REGISTRATION_STAGE or WAFER_REGISTRATION_MODE, perhaps.

drnlm:
What's the current approach for getting an overview of the pre-registration data?

docs?

stefanor:
> What's the current approach for getting an overview of the pre-registration data?

The registration Form class.

drnlm:
That just shows the user their own answers, right? There doesn't seem to be an obvious way to get an overview of everyone has pre-registered and their requests.

stefanor:
> That just shows the user their own answers, right

Yep. I can add a CSV export. But really, we need to expose KV on the API.

wafer/settings.py
```diff
@@ -259,6 +259,13 @@
 
 # Set this to False to disable registration
 WAFER_REGISTRATION_OPEN = True
+# Can be 'ticket' for Quicket tickets or 'form' for a classic form
+WAFER_REGISTRATION_MODE = 'ticket'
+
+# For REGISTRATION_MODE == 'form': The form to present, and the Group to
+# associate the responses with in the KV store
+WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
+WAFER_REGISTRATION_GROUP = 'Registration'
```
drnlm: We currently have management/commands/wafer_add_default_groups.py to easily add the page edit and talk review groups. The "Registration" group should arguably be added to that list, so it's in place out of the box.

wafer/users/forms.py
```diff
@@ -43,3 +46,37 @@ def __init__(self, *args, **kwargs):
...
         exclude = ('user', 'kv')
+
+
+def get_registration_form_class():
+    return import_string(settings.WAFER_REGISTRATION_FORM)
+
+
+class ExampleRegistrationForm(forms.Form):
+    debcamp = fields.BooleanField(
+        label='Plan to attend DebCamp', required=False)
```
drnlm: Should these be marked as translatable?

wafer/users/templates/wafer.users/registration.html
```diff
@@ -0,0 +1,7 @@
+{% extends "wafer/base.html" %}
+{% load i18n %}
+{% load crispy_forms_tags %}
+{% block content %}
+<h1>{% trans 'Registration:' %}</h1>
+{% crispy form %}
+{% endblock %}
```
drnlm: If the <user>/register url is going to be unconditionally accessible, this should check the status of WAFER_REGISTRATION_MODE and report something sensible for the 'ticket' case.

wafer/settings.py
stefanor: In that case I can drop this configuration option.

wafer/users/forms.py
```diff
@@ -43,3 +46,37 @@ def __init__(self, *args, **kwargs):
...
+    debconf = fields.BooleanField(
+        label=_('Plan to attend DebConf'), required=False)
+    require_sponsorship = fields.BooleanField(
+        label=_('Will require sponsorship'), required=False)
+
+    def __init__(self, *args, **kwargs):
+        super(ExampleRegistrationForm, self).__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.layout = Layout(
+            Fieldset('Pre-Registration',
```
drnlm: "Pre-Registration" should probably also be marked translatable.

drnlm:
I'd still like to seem some docs added about this, but we can probably land this and fix things up later

hodgestar:
:+1:

PR #205, opened 2016-03-05
Sponsor list rendering on arbitrary templates

stefanor:
e.g. the base template

This lets you put sponsor lists on every page, quite easily.

I went with a template tag, rather than a global context addition, so that it can be included anywhere, trivially.


hodgestar:
The `logo` file is a bit magic, but the file attachments need a lot of improvement anyway, so I'm :+1:.

stefanor:
Yeah, it is. I don't really have a better idea, there, though :(

PR #206, opened 2016-03-09
Use name-string interpolation to make django-admin makemessages happy

drnlm:
django-admin makemessages dislikes positional interpolation (as discussed in https://docs.djangoproject.com/en/1.9/topics/i18n/translation ), so we should keep it happy.


hodgestar:
:+1:

PR #207, opened 2016-03-11
Fix (and simplify) the DEBUG media server

stefanor:
It needs to come before the catchall in the URL router.


stefanor:
@whodafly: This is the problem you were running into

hodgestar:
:+1:

PR #209, opened 2016-03-12
Participant privacy

### LINKED ISSUE #169: Limiting the list of associated authors
It's great that Wafer lets me add additional authors to a talk. However, this means that currently anyone submitting a talk has access to the list of registered users (and no way to specify co-authors not yet registered). While the latter is only a nuisance and requires later editing, the former has been considered a privacy problem in the DebConf context.

Would you entertain an alternative approach? This could be two-fold:

1. Provide a text-field allowing the free-form addition of additional authors. Entries here could be matched against the username of the Users table to give access to augmented information (e.g. show the full name), but could also just be displayed as-is in case there is no match. This can be abused and might yield orphan data, but the information is purely informational anyway.

2. Allow users to "claim talks", causing them to show up in the list of possible authors for a talk, from which then the author can select to confirm co-authorship. This is a bit more complex, but IMHO the right way to do things.

stefanor:
Fixes #169 

DebConf would prefer to not make the list of registered users public. Talk admins can set alternate speakers, where necessary.

### DIFF
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -263,6 +263,9 @@ WAFER_REGISTRATION_MODE = 'ticket'
 # For REGISTRATION_MODE == 'form', the form to present
 WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
 
+# Allow registered and anonymous users to see registered users
+WAFER_PUBLIC_ATTENDEE_LIST = False
+
 # Ticket registration with Quicket
 # WAFER_TICKET_SECRET = "i'm a shared secret"
 
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -267,7 +267,7 @@ WAFER_REGISTRATION_MODE = 'ticket'
 WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
 
 # Allow registered and anonymous users to see registered users
-WAFER_PUBLIC_ATTENDEE_LIST = False
+WAFER_PUBLIC_ATTENDEE_LIST = True
 
 # Ticket registration with Quicket
 # WAFER_TICKET_SECRET = "i'm a shared secret"
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -152,8 +152,6 @@ INSTALLED_APPS = (
     'registration',
 )
 
-from django.db import migrations
-
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
 
 # A sample logging configuration. The only tangible logging
--- a/wafer/talks/forms.py
+++ b/wafer/talks/forms.py
@@ -14,6 +14,7 @@ from wafer.talks.models import Talk, render_author
 class TalkForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop('user')
+        kwargs.setdefault('initial', {}).setdefault('authors', [self.user])
         super(TalkForm, self).__init__(*args, **kwargs)
 
         if not self.user.has_perm('talks.edit_private_notes'):
--- a/wafer/talks/forms.py
+++ b/wafer/talks/forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.conf import settings
 from django.core.urlresolvers import reverse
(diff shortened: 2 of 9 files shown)


wafer/settings.py
```diff
@@ -265,6 +263,9 @@
 # For REGISTRATION_MODE == 'form', the form to present
 WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
 
+# Allow registered and anonymous users to see registered users
+WAFER_PUBLIC_ATTENDEE_LIST = False
```
drnlm: Do we really want this to default to False? Keeping the list private doesn't strike me as what most people will want

drnlm:
The user list isn't linked from anywhere on the site by default, and people can still play guessing games to access the indivual profile pages with this approach, so, other than making it harder to select additional authors, I'm struggling to see the significant gain here.

stefanor:
> The user list isn't linked from anywhere on the site by default

Code is open, though.

> people can still play guessing games to access the individual profile pages with this approach

Good point. That's easy enough to 404, for people with unapproved talks, too

wafer/users/views.py
```diff
@@ -87,7 +96,10 @@ class RegistrationView(EditOneselfMixin, FormView):
     template_name = 'wafer.users/registration.html'
 
     def get_user(self):
-        return UserProfile.objects.get(user__username=self.kwargs['username'])
+        try:
+            return UserProfile.objects.get(user__username=self.kwargs['username'])
+        except UserProfile.DoesNotExist:
+            raise Http404
```
hodgestar: Should we be doing `raise Http404("Reason")`? And in the other views above?

wafer/users/views.py
```diff
@@ -38,6 +38,13 @@ class ProfileView(DetailView):
     slug_field = 'username'
     slug_url_kwarg = 'username'
 
+    def get_object(self, *args, **kwargs):
+        object_ = super(ProfileView, self).get_object(*args, **kwargs)
+        if not settings.WAFER_PUBLIC_ATTENDEE_LIST:
+            if not object_.userprofile.accepted_talks().exists():
+                raise Http404()
```
drnlm: Surely this is too restrictive? Logged in users should be able to see their own profile, even if they don't have an accepted talk.

wafer/talks/forms.py
```diff
@@ -20,6 +20,14 @@ def get_talk_form_class():
 class TalkForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop('user')
+        initial = kwargs.setdefault('initial', {})
+        authors = initial.setdefault('authors', [self.user])
+
+        if not (settings.WAFER_PUBLIC_ATTENDEE_LIST
+                or self.user.has_perm('talks.change_talk')):
+            self.base_fields['authors'].limit_choices_to = {
+                'id__in': [author.id for author in authors]}
```
hodgestar: Just confirming that ordinary users then cannot add anyone else to their talk? I think that's okay, just checking that other agree.

hodgestar:
Left a couple of thoughts, otherwise looks okay to me.

wafer/talks/forms.py
drnlm: If we're not going to have the user list public, I don't think we have any better option than to restrict access to the author list like this.

wafer/users/views.py
stefanor: Yeah, that's what happens when I rush a commit to get to lunch :(

wafer/users/views.py
stefanor: The point of this is not to give any reason. If we give a reason, we have to handle the privacy case specially.

wafer/talks/forms.py
stefanor: Yeah, admins will have to do it. There's no way to allow that, without having privacy problems, really.

wafer/users/views.py
hodgestar: The reason isn't shown to anyone unless the `DEBUG=True` (at least according to the Django docs -- https://docs.djangoproject.com/en/1.9/topics/http/views/#the-http404-exception).

hodgestar:
:+1: from me.

PR #210, opened 2016-03-12
Customizeable talk form

stefanor:
Similarly to the way we customized registration forms, allow replacing the talk form.

This lets conferences tweak labelling and layout.


drnlm:
:+1:

PR #211, opened 2016-03-13
Replace "Register" with "Sign up" in account creation

stefanor:
Registering for an account on the wafer site is not registering attendance for the conference. This could avoid confusion.

### DIFF
--- a/wafer/templates/wafer/nav.html
+++ b/wafer/templates/wafer/nav.html
@@ -66,7 +66,7 @@
           </li>
           {% elif not WAFER_HIDE_LOGIN %}
           <li><a href="{% url 'auth_login' %}">
-              {% trans 'Register / Log In' %}
+              {% trans 'Sign up / Log In' %}
           </a></li>
           {% endif %}
         </ul>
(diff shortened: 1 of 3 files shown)


wafer/templates/wafer/nav.html
```diff
@@ -66,7 +66,7 @@
           </li>
           {% elif not WAFER_HIDE_LOGIN %}
           <li><a href="{% url 'auth_login' %}">
-              {% trans 'Register / Log In' %}
+              {% trans 'Sign up / Log In' %}
```
stefanor: Maybe we should leave this one, to hint that registration lies through here?

hodgestar:
:+1:

wafer/templates/wafer/nav.html
drnlm: Given the many different registration schemes possible (PyCon ZA doesn't require an account for conference registration, only for talk submission, Debconf does something different and other conferences will be different yet again), trying to sneak hints into this space isn't going to work.

Make things consistent and detailed instructions / requirements will need to be added as a page on the site.

PR #212, opened 2016-03-13
Allow the registration form to provide some initial values

stefanor:
I want to be able to use things like the username, and user's name as default values for fields in the DebConf registration from.


hodgestar:
:+1:

Issue #213, opened 2016-03-13
Improve sponsor metadata


stefanor:
Ideally sponsors should have:
1. An ordering field. The conference may want to list sponsors in the order they signed up, or sponsors in the same package, may have paid different amounts.
2. A URL for for a link back to them.
3. named association with files. So, that e.g. a large and small logo can be associated, and the templates can find them.

There's currently a hack around 3 in wafer, which leads to multiple files with the same name (confusing in the admin) https://github.com/CTPUG/wafer/blob/299d83294c54d9c83f67bbc9e677355195a18145/wafer/sponsors/models.py#L71

DebConf16 is hacking around 2 & 3, like this:

``` html
<p>
  <a href="{{ sponsor.description | removetags:"p" }}">
    <img src="{{ sponsor.files.all.0.item.url }}">
  </a>
</p>
```

https://anonscm.debian.org/cgit/debconf-data/dc16.dc.o.git/tree/templates/wafer.sponsors/sponsors.html

hodgestar:
1 and 2 sounds like good ideas.

3 needs a broader solution that extends beyond sponsors. Two possible paths
we could take:

* I like the name idea. Perhaps we can have named files for pages and
sponsors and then a separate global assets list. We'd need to add markdown,
template and url_for support for looking up the files by (page_id, name),
(sponsor_id, name) or (global_name).

* Bigger idea is to bring in a CMS extension like Wagtail to deal with this
for us. Might be too big a hammer?

drnlm:
Do we need to support more than just images for the file list?

stefanor:
I think just having a single logo attached to each sponsor might be simpler than the file list. Do we ever need more than one image? (Let's assume the images are all SVG, so scale is not an issue)

drnlm:
We've addressed the major points of this issue.

Further improvements to the sponsor setup will be addressed in a additional issues.

PR #214, opened 2016-03-13
Define a stable ordering, even if isn't configurable

stefanor:
This is a quick hack, we should really do the things listed in #213, but this is a stop-gap solution.


hodgestar:
:+1:

PR #215, opened 2016-03-14
Parse times back into datetime objects

stefanor:


drnlm:
This looks like the sort of logic we'll probably want to use elsewhere - perhaps it should be snippets as something generically useful?

drnlm:
:+1:

PR #216, opened 2016-03-14
Give talk types a defined order

stefanor:
Again, some kind of ordering field, could be handy. (And also, restricting some talk types to admins-only, such as plenary sessions)

But this gets the job done with minimal disruption.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -28,6 +28,9 @@ class TalkType(models.Model):
     def __str__(self):
         return u'%s' % (self.name,)
 
+    class Meta:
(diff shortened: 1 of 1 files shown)


hodgestar:
Seems really odd to give talks an ordering like this. I think by title or author or time slot are all valid in different contexts. Maybe we should use title as the default one?

stefanor:
@hodgestar: These are talk types, not talks. Basically defining the order of the drop-down list, during talk submission.

hodgestar:
Ah. Doh. :+1:

PR #217, opened 2016-03-14
Make redirect_profile useful for anonymous users

stefanor:
If the user isn't logged in, we can still (eventually) send them to the profile.

This allows us to reference `/accounts/profile/` from external documentation, making the redirect useful outside wafer.

Currently we only use it as a convenient destination for forms etc.


hodgestar:
:+1:

PR #218, opened 2016-03-14
Don't use the configurable e-mail address for Debian SSO

stefanor:
Currently, users can SSO in, which creates an account. Then they change their e-mail address (e.g. from an ugly alioth one, to their canonical address) and the next time they SSO, they get a new account.

Rather, let's keep a wafer account associated with a Debian identity, for life.

Storing this in the Registration group is a bit hacky, but this information is useful to the registration team. It proves account ownership.

### DIFF
--- a/wafer/registration/sso.py
+++ b/wafer/registration/sso.py
@@ -4,10 +4,13 @@ import logging
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.db import IntegrityError
 
 import requests
+
+from wafer.kv.models import KeyValue
 
 MAX_APPEND = 20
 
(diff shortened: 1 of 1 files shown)


wafer/registration/sso.py
```diff
@@ -129,11 +133,22 @@ def debian_sso(meta):
...
+
+    user = None
+    for kv in KeyValue.objects.filter(
+            group=group, key='debian_sso_email', value=email,
+            userprofile__isnull=False):
+        if kv.userprofile_set.count() > 1:
+            message = 'Multiple accounts have Debian SSOed with address %s'
+            log.warning(message, email)
+            raise SSOError(message % email)
+        user = kv.userprofile_set.first().user
```
hodgestar: Will this raise a non-SSO error if there are no matching profiles?

stefanor: We've already enforced profiles, in the query (line 141)

hodgestar: I'm guessing that the `filter` prevents that from happening.

hodgestar:
:+1:

PR #220, opened 2016-03-16
Bugs/hookup password change

drnlm:
The password change functionality is currently not very discoverable.

This ensures we use the django-registration templates (so we have the right styling) and adds a "change password" button to the edit user page (this choice is debatable - it should possibly be on the main user page, but that already has a large number of buttons already).

We should probably also add a discoverable link to the password reset form at some stage.

### DIFF
--- a/wafer/users/templates/wafer.users/edit_user.html
+++ b/wafer/users/templates/wafer.users/edit_user.html
@@ -3,5 +3,6 @@
 {% load crispy_forms_tags %}
 {% block content %}
 <h1>{% trans 'Edit User:' %}</h1>
+<li><a href="{% url 'auth_password_change' %}" class="btn btn-default">{% trans 'Change Password' %}</a></li>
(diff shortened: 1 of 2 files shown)


hodgestar:
:+1:

wafer/users/templates/wafer.users/edit_user.html
```diff
@@ -3,5 +3,6 @@
 {% load crispy_forms_tags %}
 {% block content %}
 <h1>{% trans 'Edit User:' %}</h1>
+<li><a href="{% url 'auth_password_change' %}" class="btn btn-default">{% trans 'Change Password' %}</a></li>
```
stefanor: Should this be hidden if the user doesn't have a password? (SSO-ed in before)

hodgestar: Maybe they would like to set a password?

PR #222, opened 2016-03-17
Registration confirmation screen and e-mail

stefanor:


drnlm:
We already have a bunch of registration_* templates under the registration app - I'd prefer it if the template names here were a bit more distinct here, to help avoid later confusion.

wafer/users/templates/wafer.users/registration_confirm_mail.txt
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% blocktrans %}
+Thank you for registering for {{ conference_name }}.
```
drnlm: Why conference_name, rather than WAFER_CONFERENCE_NAME as in all the other templates?

stefanor:
> We already have a bunch of registration_* templates under the registration app

Yeah, the name was getting long, so I shortened it. Move them all to a subdir?

wafer/users/templates/wafer.users/registration_confirm_mail.txt
stefanor: Because it isn't using the same context processors as the normal templates.

I wanted to make that clearer, to anyone who is editing them.

If you look at the registration (other registration) email templates, you'll see they live in a different world, too.

drnlm: That really should be mentioned in a comment then, because just using a different name is not a great hint for people looking at things later.

drnlm:
A subdirectory would work for me.

Registration is getting used in a few different contexts, so we should think about a better set of names at some point.

wafer/users/templates/wafer.users/registration_confirm_mail.txt
stefanor: I was able to fix that \o/

drnlm:
I'm still not completely happy with the template naming, but I don't have a better suggestion, so :+1:  from me.

PR #225, opened 2016-03-18
csv module expects byte input. pre-encode everything

stefanor:


drnlm:
I'm pretty sure this is not true for python 3's csv module, so I don't think this will do the right thing when run with python 3.

hodgestar:
I think this PR might be fixing things in the wrong place. I'm guessing this has nothing to do with encodings and is rather to do with `FormRegisteredUserList.details` returning crazy things instead of a list of unicode strings?

(I don't know exactly what broke though, so I could easily be wrong).

stefanor:
@drnlm: Hrm, that's an issue. We could just yield item under python3...

@hodgestar: See the big fat note on https://docs.python.org/2/library/csv.html
We are getting unicode strings, and then trying to encode them to ascii, which doesn't work :)

drnlm:
Just  yielding under python 3 seems a reason approach.

stefanor:
Here is a far simpler approach, that works under Python2 and 3.

hodgestar:
:+1:

PR #226, opened 2016-03-29
Bugs/update dependencies drop universal wheels

drnlm:
Because of the pydns / py3dns split, we can't generate universal wheels, so drop that.

Cleanup the dependancies around markdown and django-markitup while we're here, since we no longer need to support Django 1.6.


hodgestar:
:+1:

PR #227, opened 2016-03-29
Feature/test with postgres

drnlm:
We shouldn't test against only 1 database, especially as we have generally deployed with postgres rather than sqlite.

I still need to investigate why Django 1.7 doesn't like the kv migration.

### DIFF
--- a/wafer/kv/migrations/0001_initial.py
+++ b/wafer/kv/migrations/0001_initial.py
@@ -2,13 +2,14 @@
 from __future__ import unicode_literals
 
 from django.db import migrations, models
+from django.conf import settings
 import jsonfield.fields
 
 
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('auth', '0006_require_contenttypes_0002'),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL)
     ]
 
     operations = [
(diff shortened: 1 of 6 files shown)


stefanor:
:+1: to the plan. But yes, we need to make it pass.

wafer/kv/migrations/0001_initial.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import jsonfield.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('auth', '0006_require_contenttypes_0002'),
```
hodgestar: Maybe write this as `django.contrib.auth`?

hodgestar: Or maybe `migrations.swappable_dependency(settings.AUTH_USER_MODEL)`?

hodgestar:
We do now run a lot of different builds -- too many yet?

drnlm:
It's debatable how much benefit there is to still testing with python 3.3, so we could drop those builds.

Since Django 1.7 has reached end of life, there is also an argument that we shouldn't bother with it anymore.

wafer/kv/migrations/0001_initial.py
drnlm: Thanks for pointing out migrations.swappable_dependency - that saved me a lot of time digging through the migration documentation.

hodgestar:
:+1:

PR #228, opened 2016-04-04
Order sponsors by ID on the sponsors page, too

stefanor:
#214 ordered the default case, but not this, specifically-ordered case.


hodgestar:
:+1:

PR #229, opened 2016-04-05
Fix endif tag

drnlm:
As seen recently on za.pycon.org

```
Internal Server Error: /schedule/pentabarf.xml
Traceback (most recent call last):
  File "/home/pyconza/pyconza2016/ve/lib/python3.4/site-packages/django/template/base.py", line 338, in parse
    compile_func = self.tags[command]
KeyError: 'endfor'
```

This fixes the corresponding template typo.


hodgestar:
:+1:

PR #230, opened 2016-04-07
Bugs/migrations for django 1.9

drnlm:
The use of both order_with_respect_to and ordering in wafer.schedule.Slot causes installation with Django 1.9 to fail. This opts for the reverse of #190 , dropping order_with_respect_to in favour of ordering

The other 2 migrations aren't required, but as Django consistently insists on creating them, it seems best to include them as well.


hodgestar:
:+1:

stefanor:
:+1:

PR #232, opened 2016-04-08
Add a trailing / to the page URL

stefanor:
So that APPEND_SLASH is able to work. If it thinks that any URL, without
a slash, is resolveable, it won't ever redirect.

### DIFF
--- a/wafer/pages/tests/test_pages.py
+++ b/wafer/pages/tests/test_pages.py
@@ -48,18 +48,18 @@ class PageEditTests(TestCase):
         c = Client()
         # Test without edit permission
         c.login(username=no_edit_user.username, password='aaaa')
-        response = c.get('/test_edit')
+        response = c.get('/test_edit/')
         templates = [x.name for x in response.templates]
         self.assertTrue('wafer.pages/page.html' in templates)
-        response = c.get('/test_edit', {'edit': ''})
+        response = c.get('/test_edit/', {'edit': ''})
         self.assertEqual(response.status_code, 403)
         c.logout()
         # Test with edit permission
         c.login(username=edit_user.username, password='aaaa')
-        response = c.get('/test_edit')
+        response = c.get('/test_edit/')
         templates = [x.name for x in response.templates]
         self.assertTrue('wafer.pages/page.html' in templates)
-        response = c.get('/test_edit', {'edit': ''})
+        response = c.get('/test_edit/', {'edit': ''})
         templates = [x.name for x in response.templates]
         self.assertTrue('wafer.pages/page_form.html' in templates)
         self.assertEqual(response.status_code, 200)
--- a/wafer/pages/urls.py
+++ b/wafer/pages/urls.py
@@ -11,7 +11,7 @@ router.register(r'pages', PageViewSet)
 urlpatterns = patterns(
     'wafer.pages.views',
     url(r'^api/', include(router.urls)),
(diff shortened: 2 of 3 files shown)


hodgestar:
What is APPEND_SLASH?

drnlm:
We currently make some assumptions about the urls in the static site generator. Have you checked that this doesn't break anything there?

stefanor:
> What is APPEND_SLASH?

https://docs.djangoproject.com/en/1.9/ref/settings/#append-slash

If a view has a URL with a trailing slash, it redirects requests that miss that slash to the right URL.

hodgestar:
I'm a bit -1 on this -- pages are individual resources, not collections. I don't really understand what's driving this PR though.

stefanor:
> Have you checked that this doesn't break anything there?

LGTM (with #234)

stefanor:
> I don't really understand what's driving this PR though.

```
18:48 < h01ger> https://debconf16.debconf.org/accounts/profile -> not found
18:48 < h01ger> thats the first url in the mail…
18:49 < gwolf> right - My profile is /users/gwolf
18:52 < larjona> grr!
18:53 < larjona> I copied the link from the registration open mail! 
                 https://lists.debian.org/debian-devel-announce/2016/03/msg00008.html
18:53 < larjona> what happened?
18:53 < tumbleweed> h01ger: worksforme
18:54 < tumbleweed> it should go to your profile page if you're logged in, and the login screen if you aren't
18:54 < tumbleweed> oh, trailing /
18:54 < tumbleweed> larjona: the one in the e-mail had a trailing /
18:54 < tumbleweed> easy enough to fix with a redirect...
```

Then I dived down the rabbit-hole of why Django wasn't automatically doing the redirect, like it was supposed to. And discovered that our page url just thinks *everything* is route-able, so `APPEND_SLASH` doesn't do its thing.

hodgestar:
Just to be sure I'm understanding correctly, the cost of this is that every page URL without a slash now generates a redirect?

In that case we should also update `get_absolute_url` on the Page model.

Not sure how nicely this will play with static sites -- will they now have to generate `foo/index.html` instead of `foo` for every page?

We should have namespaced pages at the start when I complained. :P

hodgestar:
It looks like `get_absolute_url` on the Page model already does the right thing (i.e. includes the trailing slash because it calls `reverse`).

hodgestar:
I'd still like to understand the consequences for the generated site, but otherwise I'm reluctantly :+1:.

stefanor:
> Not sure how nicely this will play with static sites -- will they now have to generate foo/index.html instead of foo for every page?

Yep.


We had a discussion about this in #wafer. I think @hodgestar brought me around to the idea of no trailing slashes anywhere. But that seems like a far harder change to make (especially if we already have live sites). I have a vague recollection that we liked trailing slashes for static site generation. Am I misremembering?

So, I still think this is the right thing to do, given our current propensity for `/` suffixed URLs. But that we should revisit this whole thing for the next major wafer release.

@drnlm?

drnlm:
Trailing slashes can be useful when dealing with page hierarchies, since django-medusa will nicely create the directory structure without extra work. Given that we're using trailing slashes in a number of places elsewhere, and that APPEND_SLASHES defaults to True, I think this change is OK.

drnlm:
So :+1: from me as well

PR #233, opened 2016-04-08
Copy base_fields before mutating

stefanor:
base_fields is an attribute on the class, not instances, so mutations to
its contents are permanent.

Usually, one isn't supposed to mutate base_fields, and rather mutate
fields. However, limit_choices_to is applied during ModelForm's
__init__, before we can see fields.

### DIFF
--- a/wafer/talks/forms.py
+++ b/wafer/talks/forms.py
@@ -1,3 +1,5 @@
+import copy
+
 from django import forms
 from django.conf import settings
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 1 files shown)


hodgestar:
Should we add a comment about this in the code?

drnlm:
:+1:

stefanor:
> Should we add a comment about this in the code?

Yeah, I got half way through doing that, when I made the change. And then backed down because a copy before super().__init__() means something like this is going on...

hodgestar:
Indeed, but not everyone (e.g. me) knows off-hand that `base_fields` is a class variable. Maybe just a `# copy base_fields because it's a shared class attribute`?

PR #234, opened 2016-04-08
The static site renderer needs to know about the WAFER_PUBLIC_ATTENDEE_LIST behaviour as well

drnlm:
The static site renderer should only try and render publically accessible users, otherwise things will fail horribly.


hodgestar:
:+1:

PR #235, opened 2016-04-11
Features/kv rest api

drnlm:
This provides a REST api for the KeyValue stuff.

I think this captures the correct set of permission restrictions, although I may have missed some requirements.

### DIFF
--- a/wafer/kv/serializers.py
+++ b/wafer/kv/serializers.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import PermissionDenied
 from rest_framework import serializers
 from wafer.kv.models import KeyValue
 
@@ -12,8 +13,13 @@ class KeyValueSerializer(serializers.ModelSerializer):
     # See the DRF meta-issue https://github.com/tomchristie/django-rest-framework/issues/1985
     # and various related subdisscussions, such as https://github.com/tomchristie/django-rest-framework/issues/2292
     def __init__(self, *args, **kwargs):
-        # XXX: Breaks introspection tools which call serializers
-        # without a request
+        # Explicitly fail with a hopefully informative error message
+        # if there is no request. This is for introspection tools which
+        # call serializers without a request
+        if 'request' not in kwargs['context']:
+            raise PermissionDenied("No request information provided."
+                                   "The KeyValue API isn't available without "
+                                   "an authorized user")
         user = kwargs['context']['request'].user
 
         # Limit to groups shown to those we're a member of
--- a/wafer/kv/serializers.py
+++ b/wafer/kv/serializers.py
@@ -16,9 +16,8 @@ class KeyValueSerializer(serializers.ModelSerializer):
         # without a request
         user = kwargs['context']['request'].user
 
-        if not user.is_superuser:
-            # Limit to groups shown to those we're a member of
-            groups = self.fields['group']
-            groups.queryset = user.groups
+        # Limit to groups shown to those we're a member of
+        groups = self.fields['group']
+        groups.queryset = user.groups
 
         super(KeyValueSerializer, self).__init__(*args, **kwargs)
--- a/wafer/kv/serializers.py
+++ b/wafer/kv/serializers.py
@@ -6,3 +6,19 @@ class KeyValueSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = KeyValue
+
+    # There doesn't seem to be a better way of handling the problem
+    # of filtering the groups.
+    # See the DRF meta-issue https://github.com/tomchristie/django-rest-framework/issues/1985
+    # and various related subdisscussions, such as https://github.com/tomchristie/django-rest-framework/issues/2292
+    def __init__(self, *args, **kwargs):
+        # XXX: Breaks introspection tools which call serializers
+        # without a request
+        user = kwargs['context']['request'].user
+
+        if not user.is_superuser:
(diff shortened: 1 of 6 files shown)


wafer/kv/serializers.py
```diff
@@ -0,0 +1,23 @@
...
+    class Meta:
+        model = KeyValue
+
+    # There doesn't seem to be a better way of handling the problem
+    # of filtering the groups.
+    # See the DRF meta-issue https://github.com/tomchristie/django-rest-framework/issues/1985
+    # and various related subdisscussions, such as https://github.com/tomchristie/django-rest-framework/issues/2292
+    def __init__(self, *args, **kwargs):
+        # XXX: Breaks introspection tools which call serializers
+        # without a request
```
hodgestar: Would it be an improvement to return an empty queryset here instead?

wafer/kv/tests/test_kv_api.py
```diff
@@ -0,0 +1,213 @@
...
+from wafer.kv.models import KeyValue
+import json
+
+
+def get_group(group):
+    return Group.objects.get(name=group)
+
+
+def create_group(group):
+    grp = Group(name=group)
```
hodgestar: We might be able to write this as `return Group.objects.create(name=group)`?

wafer/kv/tests/test_kv_api.py
```diff
@@ -0,0 +1,213 @@
...
+    for the_group in groups:
+        grp = get_group(the_group)
+        user.groups.add(grp)
+    user.save()
+    return user
+
+
+def create_kv_pair(name, value, group):
+    group = get_group(group)
+    kv = KeyValue(key=name, value=value, group=group)
```
hodgestar: Similar comment about perhaps writing this as `return KeyValue.objects.create(...)`.

wafer/kv/tests/test_kv_api.py
```diff
@@ -0,0 +1,213 @@
...
+        self.kv_2_grp1 = create_kv_pair("Val 1.2", '{"key3": "True"}', "group_1")
+        self.kv_3_grp1 = create_kv_pair("Val 1.3", '{"key1": "Data"}', "group_1")
+        self.kv_3_grp3 = create_kv_pair("Val 3.1", '{"key1": "Data"}', "group_3")
+        self.duplicate_name = create_kv_pair("Val 1.1", '{"key1": "Data"}', "group_3")
+        self.client = Client()
+
+    def test_unauthorized_users(self):
+        response = self.client.get('/kv/api/kv/')
+        self.assertEqual(response.data['count'], 0)
+        response = self.client.get('/talks/api/kv/%d/' % self.kv_1_grp1.pk)
```
hodgestar: Is the URL meant to start with `/talks/`?

wafer/kv/serializers.py
drnlm: I'm not at all certain. The introspection issue is mentioned in the discussions on the django rest framework issues, but I haven't looked into the tools mentioned, and I'm not sure if they're actually something we need to be concerned about in wafer.

hodgestar: Maybe we should add a permission requiring that people be logged in in order to use the KV store API?

hodgestar:
Left a few questions, but generally looks good to me.

wafer/kv/tests/test_kv_api.py
drnlm: Err. Not really, no. I will fix that.

wafer/kv/serializers.py
drnlm: The Group Permission should restrict us to only logged in users already, as someone who isn't logged in doesn't belong to any groups.

The issue with the introspection case is it breaks the assumption that kwargs['context']['request'] is defined, since that is only try if the serializer is called from a proper view. While it's simple enough to return an empty queryset in that case, it's not clear to me whether that's the right thing to do to get sensible results from the introspection tools, which is why I haven't gone that route.

hodgestar: Perhap let's just raise an explicit error then that explains the situation?
Then anyone who triggers it will know what to do.

stefanor:
This gives access to the raw KV pairs, but without knowing what objects they are associated with, is it useful?

hodgestar:
@stefanor Eep. Good point.

drnlm:
We already expose the keyvalue pairs in other apis, such as for talks. While it's probably a useful extension to also list the associated objects when querying the kv objects directly, I don't see that it's required to make the api useful. And this is an improvement on the current state where KV objects cannot be easily manipulated via api calls.

For example

```
$ curl -u "<user>" http://127.0.0.1:8000/talks/api/talks/27/ | json_pp

{
   "kv" : [
      2
   ],
   "talk_id" : 27,
   "corresponding_author" : 2,
   "abstract" : "<p>This</p>",
   "status" : "P",
   "authors" : [
      2
   ],
   "title" : "Talk 2",
   "talk_type" : 1
}

$ curl -u "<user>" http://127.0.0.1:8000/kv/api/kv/2/ | json_pp

{
   "value" : "Value",
   "id" : 2,
   "group" : 3,
   "key" : "test"
}
```

stefanor:
Ah, I see. Yeah, that interface is icky. But it is workable.

stefanor:
:+1: this seems consistent

Issue #237, opened 2016-04-13
Navigation menu can show the wrong user name


drnlm:
The default navigation template sets the user details shown from the "user" object passed through the template. For viewing other user accounts, though, context["user"] is set to the user account being viewed (because detail and update view involved automatically create a context entry for the object using get_context_object_name, which happens to be user in this case), so the side-bar menu shows the wrong name.

I don't think this has any security implications, since the checks that matter are using the correct object, but the behaviour is somewhat surprising to say the least.

drnlm:
Closed by #238

PR #238, opened 2016-04-13
Add context_object_name to user views so we don't have multiple clashing meanings for 'user' object in the templates

drnlm:
This is the obvious fix for #237 

I haven't audited all the views to see if there are other situations this error can occur, though.


hodgestar:
:+1:

PR #239, opened 2016-04-14
Only show the change password button on the user's edit password page

drnlm:
An iteraction I missed in #238 

It isn't particularly sensible to show a "change password" button on every edit user screen. The change password form only changes the logged in user's password, but when shown on every screen, it appears to be an option to change the password for any user with a password. This only affects people with suitable edit user permissions, since ordinary users can't access the 'edit user' page of other people, but it is still not great UI.


hodgestar:
:+1:

PR #241, opened 2016-04-18
initial isn't yet, when editing an existing talk

stefanor:
Populate it ourselves, if necessary. Otherwise, we'll have the effect of
clobbering the authors list whenever we edit a talk.


hodgestar:
:+1:

PR #243, opened 2016-05-14
Add required decode for python 3 support

### LINKED ISSUE #242: tickets should be decoded on python 3
As seen from the recent quicket hook posts

TypeError at /tickets/quicket_hook/
the JSON object must be str, not 'bytes'

drnlm:
Should close #242

### DIFF
--- a/wafer/tickets/views.py
+++ b/wafer/tickets/views.py
@@ -75,7 +75,8 @@ def quicket_hook(request):
     if request.GET.get('secret') != settings.WAFER_TICKETS_SECRET:
         raise PermissionDenied('Incorrect secret')
 
-    payload = json.load(request)
+    # This is required for python 3, and in thweory fine on python 2
+    payload = json.load(request.decode('utf8'))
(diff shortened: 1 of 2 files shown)


wafer/tickets/views.py
```diff
@@ -75,7 +75,8 @@ def quicket_hook(request):
     if request.GET.get('secret') != settings.WAFER_TICKETS_SECRET:
         raise PermissionDenied('Incorrect secret')
 
-    payload = json.load(request)
+    # This is required for python 3, and in thweory fine on python 2
```
hodgestar: s/thw/th/ :)

hodgestar:
Other than minor typo in comment, looks good to me. :)

hodgestar:
:+1: if this has been tested somehow.

drnlm:
I've added a test case, and we've ow had one successful post from quicket on the za.pycon.org site, so I believe the fix is now correct.

hodgestar:
:+1:

PR #245, opened 2016-05-14
Forgotten schedule editor changes from November last year

zoidbergwill:
We were originally trying to delete with the page/talk ID, instead of the schedule item ID.

### DIFF
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -169,7 +169,6 @@ def check_schedule():
     return True
 
 
-@cache_result('wafer_schedule_validate_schedule', 30)
 def validate_schedule():
     """Helper routine to easily test if the schedule is valid"""
     all_items = prefetch_schedule_items()
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -169,6 +169,26 @@ def check_schedule():
     return True
 
 
+@cache_result('wafer_schedule_validate_schedule', 30)
+def validate_schedule():
+    """Helper routine to easily test if the schedule is valid"""
+    all_items = prefetch_schedule_items()
+    errors = []
+    if find_clashes(all_items):
+        errors.append('Clashes found in schedule.')
+    if find_duplicate_schedule_items(all_items):
+        errors.append('Duplicate schedule items found in schedule.')
+    if validate_items(all_items):
+        errors.append('Invalid schedule items found in schedule.')
+    if find_overlapping_slots():
+        errors.append('Overlapping slots found in schedule.')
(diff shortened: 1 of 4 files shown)


wafer/schedule/admin.py
```diff
@@ -169,6 +169,26 @@ def check_schedule():
...
+        errors.append('Invalid schedule items found in schedule.')
+    if find_overlapping_slots():
+        errors.append('Overlapping slots found in schedule.')
+    if find_non_contiguous(all_items):
+        errors.append('Non contiguous slots found in schedule.')
+    if find_invalid_venues(all_items):
+        errors.append('Invalid venues found in schedule.')
+    return errors
+
+
```
drnlm: There's benefit to consolidating valiate_schedule and check_schedule, but we shouldn't blindly replace one with the other. check_schedule is called when rendering the schedule, so we want to cache it's result for a long time as the schedule view is accessed fairly often during the conference. This is quite different from the caching behaviour we want for validation during editing, when responsiveness is more of an issue abnd it's not being accessed by many people at once.

zoidbergwill: This was why I changed the caching period for this validation to a lot lower, but I guess we probably don't wanna cache it at all.

I considered refactoring `check_schedule` to call `validate_schedule`, but we want `check_schedule` to exit early, while `validate_schedule`, but we could have that as an `arg`/`kwarg` to `validate_schedule`, because I think multiple validation errors are useful in the editor, much like Django form validation works.

drnlm:
Having played with this branch a bit now, we definitely shouldn't cache the results of ```validate_schedule```.  The lack of immediate feedback when introducing an error is going to be problematic, especially when making several changes to the schedule as it introduces a serious disconnect between the action creating the error and it being reported.

It would also be nice to update the 'Unassigned Talks' list immediately after assigning or deleting a talk - currently one has to skip between days to refresh those, which is a bit annoying.

zoidbergwill:
Sorry, I missed your comment some how, and only realised when Stefanor pinged me.

I've dropped caching the validation, and will try do the unassigned talks list ASAP.

zoidbergwill:
> It would also be nice to update the 'Unassigned Talks' list immediately after assigning or deleting a talk - currently one has to skip between days to refresh those, which is a bit annoying.

This does sound annoying. Unfortunately I'm not too sure when I'll be able to get to this. I'll try get to it ASAP, probably after Tuesday next week.

drnlm:
While it still doesn't work as nicely as I'd like, the fixed deletion is a definite improvement, so I think we can merge this as is and improve it further in another pull request.

PR #249, opened 2016-05-23
Bugs/issue 247 avoid iteritems

### LINKED ISSUE #247: Fix Python 3 regressions
There are a couple places we use `.iteritems()` on dictionaries in the code:

https://github.com/CTPUG/wafer/blob/a120735a20ff67c6f863c6ca89cc6dfd9b0fc16e/wafer/schedule/templates/admin/scheduleitem_list.html#L14

https://github.com/CTPUG/wafer/blob/a120735a20ff67c6f863c6ca89cc6dfd9b0fc16e/wafer/schedule/templates/admin/scheduleitem_list.html#L50

https://github.com/CTPUG/wafer/blob/a120735a20ff67c6f863c6ca89cc6dfd9b0fc16e/wafer/registration/sso.py#L69

https://github.com/CTPUG/wafer/blob/2dbf7020ef0c7f7d7541b68b1423fcbe5cb8ad8f/wafer/users/views.py#L173

Since `six` is bundled with Django, we can probably use that instead of using `.items()` across the board because of the performance issues on Python 2.

I'm not sure what the simplest solution for the `iteritems` uses in the templates though.

drnlm:
Closes #247 by replacing iteritems with items.

None of the mentioned uses of iteritems are performance critical, so I don't see any need for a more complicated fix.

### DIFF
--- a/wafer/registration/sso.py
+++ b/wafer/registration/sso.py
@@ -66,7 +66,7 @@ def _configure_user(user, name, email, profile_fields):
 
     profile = user.userprofile
     if profile_fields:
-        for k, v in profile_fields.iteritems():
+        for k, v in profile_fields.items():
             setattr(profile, k, v)
     profile.save()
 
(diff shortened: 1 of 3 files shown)


hodgestar:
:+1:

PR #254, opened 2016-06-08
Kv api provide links to related objects

drnlm:
As discussed with the initial kv api, this adds the related objects, and provides links to the relevant api calls when applicable.

This also handles some special cases around permissions a bit better.


drnlm:
This breaks on Django 1.7, because of the different _meta API there. Given 1.7's age, I'm not convinced fixing this is worth the hassle.

wafer/kv/serializers.py
```diff
@@ -1,8 +1,57 @@
 from django.core.exceptions import PermissionDenied
+from django.core.urlresolvers import NoReverseMatch
 from rest_framework import serializers
 from wafer.kv.models import KeyValue
 
 
+class MaybeHyperlinkField(serializers.HyperlinkedRelatedField):
+    """Overrider HyperlinkRelatedFiled to return just the object name
```
hodgestar: s/Overrider/Override/

wafer/kv/serializers.py
```diff
@@ -1,8 +1,57 @@
...
+            return cls.many_init(*args, view_name='keyvalue-detail',
+                                 queryset=KeyValue.objects.none(), **kwargs)
+        return super(CustomKeyValueSerializer, cls).__new__(cls, *args,
+                                                            **kwargs)
+
+    def get_queryset(self):
+        """Custom queryset to limit key-value pairs returned to the
+           requesting user's group for choices"""
+        request = self.context.get('request', None)
+        if request.user.id is not None:
```
hodgestar: Should this be `if request and request.user.id is not None:` to handle the case where there is no request?

hodgestar:
Could we get some tests so that I get a feel for how this works in practice?

wafer/kv/serializers.py
```diff
@@ -1,8 +1,57 @@
...
+
+    def get_url(self, obj, view_name, request, format):
+        # Because of how rest-framework creates the list for the existing values
+        # without evaluating get_queryset, we redact the non-group elements here
+        # This does leak information about the number of KeyValue pairs set by other
+        # groups, but is simpler than removing the entries entirely.
+        if request.user.id is not None:
+            grp_ids = [x.id for x in request.user.groups.all()]
+            if obj.group.id not in grp_ids:
+                return u"hidden"
```
hodgestar: If `request.user.id is None` should we be returning `u"hidden"`?

wafer/kv/serializers.py
```diff
@@ -26,4 +75,13 @@ def __init__(self, *args, **kwargs):
         groups = self.fields['group']
         groups.queryset = user.groups
 
+        # Add fields for the models we have ManyToMany relationships defined for
+        for field in self.Meta.model._meta.get_fields():
+            if field.name not in self.fields and field.many_to_many:
+                # XXX: Is there a better way to get the correct view_name out of the url config?
```
hodgestar: Would `field.related_name` be suitable for `view_name` here?

hodgestar:
Otherwise looks good to me.

wafer/kv/serializers.py
drnlm: Unfortunately not - this needs to be the detail view name that rest_framework constructs, and that isn't a direct attribute of the model or the field.

Lookingat it again, rest_framework does provide a utility function which constructs this name, so I've switched to using that.

wafer/kv/serializers.py
drnlm: Probably a good idea.

hodgestar:
If probably makes sense to rather use nested REST endpoints as done for the talk URLs in #316 than to take the more complicated approach used here.

hodgestar:
Should we close this PR?

drnlm:
We can, although this is still useful if we have a use case for the KV stuff.

If we don't, we should also strip out the KV support.

drnlm:
No-one seems to have a compelling use case for this, and we're not using the KV stuff that much anyway, so closing.

PR #255, opened 2016-06-09
Bug/can claim should require login

drnlm:
As googlebot recently demonstrated, the claim tickets form doesn't handle the case of a not logged in user elegantly. It also doesn't make a great deal of sense to have that form accessible to anonymous users, so require login for claiming tickets.


hodgestar:
:+1:

PR #257, opened 2016-06-14
Day ID is a useless navigation indicator

zoidbergwill:
It's order is useless for navigation, while actual date is a lot
more useful.

@stefanor was complaining about this.


hodgestar:
:+1:

PR #259, opened 2016-06-14
Sparse schedule support

stefanor:
If you want a schedule item to span an entire row, it must be the only
thing in the row, and have the "expand" css_class.

### DIFF
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -186,7 +186,8 @@ class ScheduleItemAdmin(admin.ModelAdmin):
     form = ScheduleItemAdminForm
 
     change_list_template = 'admin/scheduleitem_list.html'
-    list_display = ['get_start_time', 'venue', 'get_title', 'expand']
+    list_display = ('get_start_time', 'venue', 'get_title', 'expand')
+    list_editable = ('expand',)
 
     # We stuff these validation results into the view, rather than
     # enforcing conditions on the actual model, since it can be hard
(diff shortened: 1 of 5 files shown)


stefanor:
The DebConf16 schedule is rather sparse, and thus a little misleading.

Before:
before

After:
after

drnlm:
"The only item in a row" is not a restriction we should require. It breaks use cases with venues that may be split or combined. Scheduling a talk in the combined venue while a tutorial is running in parallel is hardly an uncommon use case, and one we still want to support.

stefanor:
> Scheduling a talk in the combined venue while a tutorial is running in parallel is hardly an uncommon use case, and one we still want to support.

I don't think this is really the right mechanism for that. Currently, it expands to fill all available space.

I can remove the "only thing in the row" restriction, but it'll still just be expanding as far as possible. Quite likely into rooms that are irrelevant.

drnlm:
a) This shouldn't break the test suite.

b) This shouldn't be shoved into css_class - this is completely unrelated to that's intended use case.

stefanor:
> This shouldn't break the test suite.

Missed that because of the unrelated breakage.

> This shouldn't be shoved into css_class - this is completely unrelated to that's intended use case.

Not completely, it's a mostly visual thing. The event is still tied to a specific venue, under the hood.

drnlm:
You probably want to hookup expand as changeable on the list admin view, but that's not urgent.

:+1:

PR #260, opened 2016-06-15
Bugs/replace reversion generate patch

drnlm:
Django Reversion 2 changed it's API in a significant places, which breaks the test suite.

This unbreaks stuff for Django 1.8 and 1.9.

Unbreaking the Django 1.7 tests requires installing an older version of django-reversion, as version 2.0 no longer supports that.

### DIFF
--- a/wafer/compare/admin.py
+++ b/wafer/compare/admin.py
@@ -11,7 +11,7 @@ from django.shortcuts import get_object_or_404, render
 from django.contrib.admin.utils import unquote, quote
 from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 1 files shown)


hodgestar:
Hmm. Tests are failing.

drnlm:
Only the Django 1.7 tests are failing, and that's never going work with django-reversion 2.0.

We can either add logic to install an older version for the 1.7 tests, or drop 1.7 support (which is arguably no longer worth the effort).

hodgestar:
Hmm. I think I'm +1 on dropping 1.7 support.

drnlm:
Superceded by #280

PR #261, opened 2016-06-15
Feature/drop django 1.7 support

drnlm:
As mentioned in #260 and #254, continuing to support Django 1.7 is becoming problematic.

This drops support for Django 1.7 and upates the documenation accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,12 @@ python:
    - "3.4"
    - "3.5"
 env:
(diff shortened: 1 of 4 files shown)


stefanor:
DebConf is running on Django 1.7, can we hold this back?

drnlm:
With Debconf now over, and since the version of django-restframework that support Django 1.10 also breaks support for Django 1.7,  is there any reason not to land this?

PR #262, opened 2016-06-16
Restrict django-reversion to before 2.0

drnlm:
As a short term measure until we drop Django 1.7, restrict django-reversion to something that still supports that verison.


hodgestar:
:+1:

PR #263, opened 2016-06-16
Replace refresh meta tag with robust AJAX refresh

stefanor:
If a request fails, don't let the browser get stuck on an error screen,
rather keep trying.


hodgestar:
:+1:

PR #266, opened 2016-06-26
Expose scheduling information on talk page.

lelutin:
When ppl click on a talk name, it would be nice to see some information
about when and where it is happening. The default wafer templates do not
expose this information.

### DIFF
--- a/wafer/schedule/models.py
+++ b/wafer/schedule/models.py
@@ -209,7 +209,7 @@ class ScheduleItem(models.Model):
     def get_duration_minutes(self):
         """Return the duration in total number of minutes."""
         duration = self.get_duration()
-        return int(duration['hours']*60 + duration['minutes'])
+        return int(duration['hours'] * 60 + duration['minutes'])
 
 
 def invalidate_check_schedule(*args, **kw):
--- a/wafer/schedule/models.py
+++ b/wafer/schedule/models.py
@@ -206,6 +206,11 @@ class ScheduleItem(models.Model):
             result['hours'] += hours
         return result
 
+    def get_duration_minutes(self):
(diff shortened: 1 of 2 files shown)


hodgestar:
Looks good to me. Thanks!

One of the Travis builds failed to start, so I've restarted it.

lelutin:
about the translation blocks, I was wondering if it wouldn't be better to use a blocktrans and include the data inside the block. (e.g. start a blocktrans before `Duration:` and end the block only after `mins`, and similarly for other informations on the same page. this would make it possible to inverse text and data for languages that use a different direction.

do you think that change would make sense?

hodgestar:
@lelutin I think using `blocktrans` here makes a lot of sense.

lelutin:
ok this is hopefully my last commit for this PR. I've changed translation to use blocktrans on info lines above the talk description. I've also added `Type:` as translatable text.

wafer/schedule/models.py
```diff
@@ -206,6 +206,11 @@ def get_duration(self):
             result['hours'] += hours
         return result
 
+    def get_duration_minutes(self):
+        """Return the duration in total number of minutes."""
+        duration = self.get_duration()
+        return int(duration['hours']*60 + duration['minutes'])
```
hodgestar: I think PEP8 likes spaces around operators like (so ` * 60`).

hodgestar:
Left a minor style nitpick but the blocktrans version looks great. Tx for also updating the talk type part.

lelutin:
oops you're right about the missing space around the operator. I've pushed a fix for that

hodgestar:
:+1:

Thanks!

PR #277, opened 2016-07-11
Add simple heading callouts to sponsors list page

zoidbergwill:
To make sponsor packages more obvious.

I think I need to drop or use smaller wells, since it doesn't look great with default bootstrap.

## Default Bootstrap Theme

[Before](https://cloud.githubusercontent.com/assets/2572493/16745028/563fa572-47b4-11e6-9bc0-d5bafd1153d5.png)

[After](https://cloud.githubusercontent.com/assets/2572493/16745039/6251b08a-47b4-11e6-9b1d-e0352252d085.png)

## PyConZA Theme

(I dropped the body background colour for PyConZA's CSS to make Platinum look a bit better)

[Before](https://cloud.githubusercontent.com/assets/2572493/16745045/6df1c984-47b4-11e6-8c6c-d48d9bf45804.png)

[After](https://cloud.githubusercontent.com/assets/2572493/16745071/8394ddbc-47b4-11e6-8a48-8d64fae96507.png)

## Still to do:

- [ ] Better alignment of the actual name of the package, or dropping it, and letting platinums have stars.
- [ ] Fix well padding

### DIFF
--- a/wafer/sponsors/templates/wafer.sponsors/sponsors.html
+++ b/wafer/sponsors/templates/wafer.sponsors/sponsors.html
@@ -8,6 +8,9 @@
     <section class="wafer wafer-sponsor">
     <div class="well">
       <h2 class="bs-callout bs-callout-default bs-callout-{{ sponsor.packages.first.name.lower }}">
+         {% if sponsor.packages.first.name.lower == 'platinum' %}
+            ★
+         {% endif %}
          {% if sponsor.url %}
          <a href="{{ sponsor.url }}">{{ sponsor.name }}</a>
          {% else %}
--- a/wafer/sponsors/templates/wafer.sponsors/sponsors.html
+++ b/wafer/sponsors/templates/wafer.sponsors/sponsors.html
@@ -6,16 +6,18 @@
 <div class="wafer list">
     {% for sponsor in sponsor_list %}
     <section class="wafer wafer-sponsor">
-    <div>
-        <h2>
-           {% if sponsor.url %}
-           <a href="{{ sponsor.url }}">{{ sponsor.name }}</a>
-           {% else %}
-           {{ sponsor.name }}
-           {% endif %}
-        </h2>
-        {{ sponsor.description.rendered|safe }}
+    <div class="well">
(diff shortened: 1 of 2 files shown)


drnlm:
I think we should keep the name of the packages here, but there does need to be more separation bewtween the sponsor name and the package name.

Stars for platinum sponsors to match the menu seems a good move.

zoidbergwill:
I think the spacing issue might be only for PyConZA, because of it's custom CSS?

drnlm:
Since we need to update the PyCon ZA stuff anyway, let's merge this and fix the spacing along with the other changes.

wafer/sponsors/templates/wafer.sponsors/sponsors.html
```diff
@@ -6,16 +6,21 @@
...
-           {% if sponsor.url %}
-           <a href="{{ sponsor.url }}">{{ sponsor.name }}</a>
-           {% else %}
-           {{ sponsor.name }}
-           {% endif %}
-        </h2>
-        {{ sponsor.description.rendered|safe }}
+    <div class="well">
+      <h2 class="bs-callout bs-callout-default bs-callout-{{ sponsor.packages.first.name.lower }}">
+         {% if sponsor.packages.first.name.lower == 'platinum' %}
```
stefanor: That's a bit hacky. Not everyone names their packages like that.

zoidbergwill: Yeah, I agree. 

I mentioned it in [my commit message](https://github.com/CTPUG/wafer/pull/277/commits/811426f2f1b1048c782c249a5895e9eb0f94d7ca)

zoidbergwill: Should we be getting the most expensive sponsor?

drnlm: We've added a order attribute to sponsor packages, so we could potentially use that. It's still rather hacky, but easily controllable via the site admin.

It might be worthwhile adding the character to display as a field to the package model, for more flexibility?

zoidbergwill: Adding a field to the model definitely sounds like the best solution.

PR #278, opened 2016-07-29
Handle talks >= 1 hour correctly

stefanor:
Bugfix for #266

Really, we should probably expose this a little more prettily from the model.


drnlm:
:+1:

(Test failures are because django rest framework has dropped support for Django 1.7)

PR #280, opened 2016-08-13
Features/bump django reversion requirement

drnlm:
Suporting both django-reversion 1.10 and 2.0 requires some complexity because of the changed APIs.  Since we no longer support Django 1.7, there's little value in support django-reversion before 2.0, so this complexity doesn't seem worth it.

This supercedes #260, which aims to support both.


hodgestar:
:+1:

PR #281, opened 2016-08-13
Add edit button to talk list in the profile

### LINKED ISSUE #279: Talk editing should be more easily discoverable
The last couple of years, we've had a number Pycon ZA speakers who've required very explicit instructions on how to edit their talk submissions.

This strongly suggests we need to make talk editing more discoverable - possibly adding edit links to the talks list on the user profile page.

drnlm:
In an effort to make talk editing more obvious and close #279 , this adds edit buttons to the pending talks list in the user profile.

### DIFF
--- a/wafer/users/templates/wafer.users/profile.html
+++ b/wafer/users/templates/wafer.users/profile.html
@@ -110,10 +110,12 @@
 {% for talk in profile.pending_talks %}
 <div class="well">
     <a href="{{ talk.get_absolute_url }}">{{ talk.title }}</a>
-    <!-- Because this is one of the author's pending talks, we don't need to
-         check for edit permission's on the talk explictly. This doesn't show
-         the edit button for people with 'change-talk' permissions, but we
-         accept that tradeoff for simplicity here. -->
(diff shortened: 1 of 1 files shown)


wafer/users/templates/wafer.users/profile.html
```diff
@@ -110,6 +110,11 @@
 {% for talk in profile.pending_talks %}
 <div class="well">
     <a href="{{ talk.get_absolute_url }}">{{ talk.title }}</a>
+    <!-- Because this is one of the author's pending talks, we don't need to
+         check for edit permission's on the talk explictly. This doesn't show
+         the edit button for people with 'change-talk' permissions, but we
+         accept that tradeoff for simplicity here. -->
```
hodgestar: This comment will show up in the generated HTML, which I'm guessing is not what we want? Should we use `{# ... #}` or `{% comment ... %} ... {% endcomment %}` instead?

hodgestar:
One comment, otherwise looks good.

hodgestar:
:+1:

Issue #251, opened 2016-05-26
Allow disabling talk submission by talk type


drnlm:
As mentioned on IRC as a feature wanted for Debconf, so we don't forget about it

Probably useful in contexts other than Debconf, for things like invited talks and such which should created  directly by the conferenc eorganisers, rather than as talk submissions.

drnlm:
Closed by #284

PR #284, opened 2016-08-13
Features/issue 251 disable specific talk types

drnlm:
To handle #251 , this adds a 'disable_submission' field to TalkType, so types can be managed individually.

I'm leaving the global enable / disable flag in place for now as that's more convienent than toggling each type.

Since I'm changing the TalkType model, I've also added an order field, since controlling the order talk types are presented to the user can be useful.


hodgestar:
:+1:

PR #285, opened 2016-08-14
Feature/zapier webhook support

drnlm:
This is as tested as I can with the test data zapier spits, and generally seems to behave as expected.

It will probably blow up in interesting ways once deployed.


hodgestar:
:+1: -- let's give it a try?

PR #286, opened 2016-08-16
Feature/add comapre view for page editors

drnlm:
This adds a rough "compare" view for page editors, so they can examine changes without needing to go through the admin site.

It only compares past versions against the current version, which I think should cover most use cases.

### DIFF
--- a/wafer/pages/templates/wafer.pages/page_compare.html
+++ b/wafer/pages/templates/wafer.pages/page_compare.html
@@ -1,0 +1,49 @@
+{% extends "wafer/base.html" %}
+
+{% load i18n %}
+{% block content %}
+
+<h1>Comparing Page</h1>
+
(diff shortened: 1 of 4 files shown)


wafer/pages/templates/wafer.pages/page_compare.html
```diff
@@ -0,0 +1,49 @@
+{% extends "wafer/base.html" %}
+
+{% load i18n %}
+{% block content %}
+
+<h1>Comparing Page</h1>
+
+{% if perms.pages.change_page %}
```
hodgestar: Do we want this check here given that there is a check in the view?

hodgestar:
Other than the one comment, looks good to me.

hodgestar:
:+1: once the builds are happy.

PR #287, opened 2016-08-17
Feature/speaker page links

drnlm:
This adds links from the speakers page view to the individual profiles, so it's easier to get to the information about individual speakers and their accepted talks.

This also sorts speakers by name, since the current ordering isn't particularly useful.


hodgestar:
:+1:

PR #290, opened 2016-08-20
Add a trailing slash to the paginated talk list

stefanor:
So that the static rendered version works.


hodgestar:
:+1: -- looks like failing builds are for Django 1.7 which I thought we dropped support for? Did this branch start from the right place / merge stuff okay?

stefanor:
I started from the version we were using for DebConf16, so yeah, expected.

Issue #289, opened 2016-08-20
Speaker list public in static site, when WAFER_PUBLIC_ATTENDEE_LIST is False


stefanor:
https://github.com/CTPUG/wafer/blob/f7f23cbdbcfff353ea83448e8b537551fb525fe2/wafer/users/renderers.py#L13-L14

Not rendering `/users/` anything when `WAFER_PUBLIC_ATTENDEE_LIST` is False means we are likely to fall back to the webserver's directory listing endpoint on `/users/`.

That will show a very ugly list of all the speakers.

medusa doesn't seem to be able to handle http exceptions, which is why we don't try to render `/users/`.

I think the best thing to do would be to just filter the user list down to the approved speaker list, rather than 404ing. Agreed?

hodgestar:
Sounds like the correct approach to me.

PR #291, opened 2016-08-21
Presenter list

stefanor:
Order `/users/` bit. And generally resuscitate it when `WAFER_PUBLIC_ATTENDEE_LIST` is False.

Fixes: #289

### DIFF
--- a/wafer/users/views.py
+++ b/wafer/users/views.py
@@ -41,6 +41,7 @@ class UsersView(ListView):
         qs = super(UsersView, self).get_queryset(*args, **kwargs)
         if not settings.WAFER_PUBLIC_ATTENDEE_LIST:
             qs = qs.filter(talks__status=ACCEPTED).distinct()
+        qs = qs.order_by('first_name', 'last_name', 'username')
         return qs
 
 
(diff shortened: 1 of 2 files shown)


wafer/users/views.py
```diff
@@ -37,9 +38,11 @@ class UsersView(ListView):
     paginate_by = 25
 
     def get_queryset(self, *args, **kwargs):
+        qs = super(UsersView, self).get_queryset(*args, **kwargs)
         if not settings.WAFER_PUBLIC_ATTENDEE_LIST:
-            raise Http404()
-        return super(UsersView, self).get_queryset(*args, **kwargs)
+            qs = qs.filter(talks__status=ACCEPTED).distinct()
```
hodgestar: I'm curious about the need for `distinct` here.

stefanor: Otherwise the speaker appears multiple times if they have multiple talks. I noticed this with my sample dc16 dataset.

hodgestar:
There is an odd test failure, but it looks unrelated so I've restarted the build. :+1: once the builds are green.

PR #294, opened 2016-09-05
Get rid of /index/ in static site

stefanor:
And copy static assets in place


hodgestar:
Looks good to me, but leaving this to @drnlm since he knows the static site generator better than me.

wafer/management/static.py
```diff
@@ -19,10 +15,3 @@ def render_path(self, path=None, view=None):
                                                                      view)
             except IOError as err:
                 print('Skiping %s - threw IOError %s' % (path, err))
-            # This is a hack because dajngo_medusa doens't understand 301
```
drnlm: This hack exists because django-medusa converts every response that isn't 200 into a fatal generic Exception.

As long as we have the RedirectView in pages/urls.py, I'm pretty sure we're going to need something to handle this case, since we'll still be encountering a redirect while walking the tree of possible urls.

stefanor: Why will we need a `/index`? Nothing links to it.

drnlm:
This ends up rendering the index page as "/index/index.html" - I think we probably need to add logic so the bare url is handled correctly.

stefanor:
> This ends up rendering the index page as "/index/index.html"

I don't get that. I just get `/index.html`. Any idea why you don't?

drnlm:
I assume you're testing against the dc16 repo.

That adds the dc16 app, which amongst other things includes it's own custom renderers.py that explictly returns '/' as a path, which is what generates /index.html.

That's probably something we want to add to wafer.

PR #305, opened 2016-10-07
m is months, not minutes. This causes confusion

drnlm:
The minutes display should not remain constant.

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/current.html
+++ b/wafer/schedule/templates/wafer.schedule/current.html
@@ -51,7 +51,7 @@
    </table>
    {% endif %}
    <div class="timestamp">
-     Last updated: {% now "H:m:s" %}
+     Last updated: {% now "H:i:s" %}
      <div class="errors"></div>
    </div>
 </div>


wafer/schedule/templates/wafer.schedule/current.html
```diff
@@ -51,7 +51,7 @@
    </table>
    {% endif %}
    <div class="timestamp">
-     Last updated: {% now "H:m:s" %}
+     Last updated: {% now "H:i:s" %}
```
hodgestar: Shouldn't this be `%M`?

drnlm: Not according to Djangos documentation for the now builtin, which refers to the filter for date (https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#std:templatefilter-date)

Quoting the relevant lines from their table:

> i	  Minutes.	  '00' to '59'
> m Month, 2 digits with leading zeros.  	'01' to '12'
> M Month, textual, 3 letters.	  'Jan'

Issue #236, opened 2016-04-12
nav.html should be easier to modify / extend


drnlm:
Both debconf (https://anonscm.debian.org/cgit/debconf-data/dc16.dc.o.git/tree/templates/wafer/nav.html) and the latest pyconza2016 facelift pull request have ended up completely copying nav.html for fundamentally minor changes. This seems fair from ideal, and we should to revisit how nav.html is structured so this isn't required.

It's not clear to me how best to address this - possibly we need to split the innermost bits (such as the loop creating the main menu list and the dropdown menu logic) into sub-templates that are included, rather than having everyone duplicate that logic.

zoidbergwill:
I guess semi-related, we want to make it easier for people to customise the menus and nav, so I guess that fits in to this ticket too.

PR #308, opened 2016-10-08
Add a bunch of blocks to nav.html

drnlm:
This adds several blocks to nav.html, to partially address #236


stefanor:
I don't think this is actually useful. There is nowhere to override these blocks.

It'd probably make more sense to put each of these blocks in a separate file, so they can be overridden like that.

PR #312, opened 2016-10-09
Clear extra fields when changing items through the schedule view

### LINKED ISSUE #300: Schedule Editor does not clear extra fields on existing items
When replacing an existing item in the schedule editor, the notes, css_class and details fields are not replaced or cleared.

While this can be useful to leave css_class untouched, it is surprising behaviour and usually the wrong thing to do for notes  and details. 

I think we must clear notes and details when this happens.

We should probably clear all extra fields when replacing an item, for the sake of predictablity,

drnlm:
Should close #300

### DIFF
--- a/wafer/schedule/serializers.py
+++ b/wafer/schedule/serializers.py
@@ -34,6 +34,12 @@ class ScheduleItemSerializer(serializers.HyperlinkedModelSerializer):
             existing_schedule_item.talk = talk
             existing_schedule_item.page = page
             existing_schedule_item.slots = slots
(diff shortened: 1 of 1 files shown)


drnlm:
This changes create, since that's hit when we drag a talk or page from the side page into the schedule, as those items don't have a data-scheduleitem-id property, which is what's checked for in edit_schedule.js .

update should be hit when dragging items from one position to another in the schedule. That doesn't currently work for we with any browser, but the default update behaviour of copying all the fields is probably correct there.

Issue #150, opened 2015-11-27
Ability to lock pages (external data import)


madduck:
Traditionally, DebConf kept web content in version control and some of us want to keep it that way, rather than requiring users to edit the webpages through the web and losing versioning along the way.

One idea that came up recently was to load data from Git into Django with a `post-receive` hook. To prevent users' edits through the web from being overwritten, editing should thus be disabled. It could be done globally (remove the "change_page" permission), or on a per-page basis, but the latter requires a "locked" field in the database (since Django permissions are per-task, not per-object).

This is low-priority but might be something I'll just hack up in a night as the actual Git export seems trivial, and I need similar functionality for another (paid) project anyway.

hodgestar:
Is the idea that the Git repository would store JSON objects that represent the entire page model contents, or just the page markdown?

hodgestar:
I'm a bit against completely locking out edits from Django admin (given that sometimes admins just need to fix things right now) but perhaps we could add an "externally maintained" flag that warns people that their edits might be overwritten by some external system? Perhaps the first implementation should just rely on updating page content via the API (#181) and we can then fiddle with the details once we have a sense of what the publishing workflow from Git looks like?

madduck:
@hodgestar: Git would hold just the page markdown, really. There are metadata that might need to be encoded in the file (e.g. title), but I wouldn't use JSON for that.

The Git workflow is simple: keep the content outside of the database as using Git allows for better collaboration, asynchronous workflow, analysis (diff, blame, etc.), and trivial reuse of content. Since wafer instances are one-per-conference, but a lot of the static text we write for the conference site essentially comes from somewhere else or will be used again in the future, shoving it in and out of a database for each year seems wrong.

At least for DebConf, I see a three stage process: (1.) the wiki collects pretty much everything in quick-n-dirty fashion; (2.) certain stuff from there gets pulled into Git, amassing a resource usable across the years; and (3.) each wafer instance has access to the static text in Git and mainly adds all the dynamic handling around it. IMHO, Django and through-the-web-editing are great for object-oriented/model-based editing workflows, but simple text is much better handled using proper editors and content trackers.

PR #314, opened 2016-10-09
Add a management command to load pages from markdown pages on disk

stefanor:
Fixes: #150. Mostly. Without locking pages, as that probably isn't
necessary for now.

Pages are layed out matching the URL structure, with jekyll-style front
matter to describe metadata.

Not declaring a pyyaml dependency, as it's only needed in this management command, and thus rather optional.

### DIFF
--- a/docs/pages.rst
+++ b/docs/pages.rst
@@ -29,3 +29,38 @@ Files
 Additional files, such as images, can be uploaded so they can be referenced in page.
 These files are placed in MEDIA_ROOT/pages_files by default. This location needs to
 be writable by the webserver for uploads to work.
+
+Maintaining pages in files
+==========================
+
+Pages live in the database, and can be edited through the web UI.
+However, it can be useful to store them in files (e.g. in a git repo,
+with the site source code).
+
+There is a management command (``load_pages``) that will read pages from
+files into the database.
+It requires ``PyYAML`` to be installed.
+
+Pages must be stored as markdown in a directory, in the same hierarchy
+as the desired URL structure.
+The ``PAGE_DIR`` Django setting should be an absolute path to root
+directory of this hierarchy, beginning and ending with ``/``.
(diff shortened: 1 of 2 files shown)


docs/pages.rst
```diff
@@ -29,3 +29,36 @@ Files
...
+
+Pages must be stored as markdown in a directory, in the same hierarchy
+as the desired URL structure.
+The ``PAGE_DIR`` setting should be an absolute path to root directory of this
+hierarchy, beginning and ending with ``/``.
+e.g.::
+
+    /app/pages/                         ← PAGE_DIR
+    /app/pages/index.md                 ← Home Page: /
+    /app/pages/about.md                 ← Container Page: /about/
```
hodgestar: It seems a bit more natural for this to be `/app/pages/about/index.md`?

docs/pages.rst
```diff
@@ -29,3 +29,36 @@ Files
...
+
+Pages live in the database, and can be edited through the web UI.
+However, it can be useful to store them in files (e.g. in a git repo,
+with the site source code).
+There is a management command (``load_pages``) that will read pages from
+files into the database.
+
+Pages must be stored as markdown in a directory, in the same hierarchy
+as the desired URL structure.
+The ``PAGE_DIR`` setting should be an absolute path to root directory of this
```
hodgestar: I would say `Django setting` here just to be clearer that this lives in `settings.py`.

docs/pages.rst
stefanor: Yes, it does. But that doesn't match how we actually render pages. The root index page is a hack. This closely mirrors the DB structure.

hodgestar: Although that doesn't match Wafer's internals that well.

hodgestar:
Looks good to me.

PR #318, opened 2016-11-09
Remove unnecessary code that breaks jQuery 3

stefanor:
event.props was removed in jquery 3 (jquery/api.jquery.com#405).

It looks like this wasn't doing anything useful, and broke the schedule
editor.

### DIFF
--- a/wafer/static/js/edit_schedule.js
+++ b/wafer/static/js/edit_schedule.js
@@ -1,8 +1,6 @@
 (function () {
     'use strict';
 
-    jQuery.event.props.push('dataTransfer');
-
     var handleDragStart = function (e) {
         e.target.style.opacity = '0.4';  // this / e.target is the source node.
         e.target.classList.add('label-danger');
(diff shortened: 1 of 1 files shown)


wafer/static/js/edit_schedule.js
```diff
@@ -65,23 +61,20 @@
...
 
-        if (e.stopPropagation) {
-            e.stopPropagation(); // stops the browser from redirecting.
-        }
+        e.preventDefault(); // stops the browser from redirecting
 
         var slot = e.target.getAttribute('data-slot');
         var venue = e.target.getAttribute('data-venue');
 
         //noinspection JSUnresolvedVariable
```
hodgestar: Can we remove this comment now?

Any idea how the `event.` references worked before? Some sort of crazy magic?

stefanor: Apparently it's a legacy thing: http://stackoverflow.com/a/20522980

hodgestar: Cool. Let's remove the `noinspection` then.

hodgestar:
:+1:

PR #328, opened 2016-11-23
Feature/better slot management

drnlm:
This adds several small improvements to the slot admin interface

* The computed day and start time is listed for all slots
* A filter to select slots by day has been aded
* A bug with adding multiple slots when using previous_slot was fixed
* I added descriptions for the admin columns in a couple of places

### DIFF
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -306,7 +306,7 @@ class SlotDayFilter(admin.SimpleListFilter):
 class SlotAdmin(admin.ModelAdmin):
     form = SlotAdminForm
 
-    list_display = ('__str__', 'get_day', 'get_start_time', 'end_time')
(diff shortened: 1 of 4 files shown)


wafer/schedule/admin.py
```diff
@@ -273,14 +273,48 @@ class SlotAdminAddForm(SlotAdminForm):
...
+    # day is dynamic (either the model field or the previous_slot)
+    title = _('Day')
+    parameter_name = 'day'
+
+    def lookups(self, request, model_admin):
+        # List filter wants the value to be a string, so we use
+        # pk to avoid bouncing through strptime.
+        return [('%d' % day.pk, str(day)) for day in Day.objects.all()]
+
+    def queryset(self, request, queryset):
```
hodgestar: I feel like we should probably have some tests for this queryset method.

hodgestar:
:+1:

PR #329, opened 2016-12-06
Features/add symbol to sponsor package

drnlm:
As discussed in #277, wafer currently uses a bit of hackery on the package name to add the '★' symbol for platinum sponsors. This is problematic for conferences using different sponsor package naming conventions, and not very flexible.

This adds an explicit 'symbol' field to make this customisable as part of the sponsor package.

This is currently restricted to single character symbols. We maybe should allow more complicated combinations here, but it's not clear to me what a good limit should be.

### DIFF
--- a/wafer/sponsors/models.py
+++ b/wafer/sponsors/models.py
@@ -39,6 +39,10 @@ class SponsorshipPackage(models.Model):
         File, related_name="packages", blank=True,
         help_text=_("Images and other files for use in"
                     " the description markdown field."))
+    symbol = models.CharField(
+        max_length=1, blank=True,
+        help_text=_("Optional symbol to display next to sponsors "
+                    "backing at this level sponsors list"))
 
     class Meta:
         ordering = ['order', '-price', 'name']
--- a/wafer/sponsors/models.py
+++ b/wafer/sponsors/models.py
@@ -39,10 +39,13 @@ class SponsorshipPackage(models.Model):
         File, related_name="packages", blank=True,
         help_text=_("Images and other files for use in"
                     " the description markdown field."))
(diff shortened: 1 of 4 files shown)


wafer/sponsors/models.py
```diff
@@ -39,6 +39,10 @@ class SponsorshipPackage(models.Model):
         File, related_name="packages", blank=True,
         help_text=_("Images and other files for use in"
                     " the description markdown field."))
+    symbol = models.CharField(
+        max_length=1, blank=True,
+        help_text=_("Optional symbol to display next to sponsors "
+                    "backing at this level sponsors list"))
```
hodgestar: The help text seems to be a bit odd. Maybe `Optional symbol to display in the sponsors list next to sponsors who have sponsored at this level, e.g. ★.`?

hodgestar: Hmm. Maybe we should make the `max_length` bigger. I can imagine people sticking short words here as a different way to mark sponsors.

drnlm: Any suggestions as what the length of the field should be?

hodgestar: Somewhere in the 64-256 range?

drnlm: I'm worried that 64 characters is too long for something that will be prefixed to the sponsor name in the list view.

drnlm:
Should this also be displayed on the packages list?

hodgestar:
+1 on also displaying it on the packages list.

wafer/sponsors/models.py
hodgestar: Eh, let's stick with 1 -- it's easier to migrate to a larger number.

PR #330, opened 2016-12-13
Bootstrap4 doesn't have navbar-right

stefanor:


wafer/templates/wafer/nav.html
```diff
@@ -14,7 +14,7 @@
     <a class="navbar-brand" href="{% url 'wafer_page' %}">{{ WAFER_CONFERENCE_NAME }}</a>
     {% endblock navtitle %}
 
-    <div class="collapse navbar-toggleable-xs pull-xs-right" id="wafer-navbar-collapse">
+    <div class="collapse navbar-toggleable-xs" id="wafer-navbar-collapse">
```
hodgestar: Just checking whether removing pull-xs-right entirely rather than replacing it with float-xs-right is the correct thing to do here.

stefanor: Yes, it pulled the whole menu to the right.

PR #333, opened 2016-12-14
More bootstrap4 nav fixes

stefanor:
#330 introduced a regression - the menu misbehaved when the hamburger was expanded. This puts the hamburger back on the left, which makes things work again. I can't figure out how to get it to work, on the right.

Also, the expand button had a double hamburger on it.


drnlm:
While this isn't perfect, it is better than the current situation, so perhaps we should merge it as is and worry about fixing the remaining niggles after that.

PR #334, opened 2016-12-24
Feature/files with names

drnlm:
As mentioned in #213 , the current assumptions requires specifying the same name for the images associated with each sponsor, which is confusing in the admin.

This changes the str representation for files added via pages and sponsors to include more information (associated objects & the path) to make it easier to distinguish between the files.

It also adds a template tag to simplify looking up images associated with a sponsor.

### DIFF
--- a/wafer/pages/models.py
+++ b/wafer/pages/models.py
@@ -18,7 +18,7 @@ from wafer.menu import MenuError, refresh_menu_cache
 class File(models.Model):
     """A file for use in page markup."""
     name = models.CharField(max_length=255)
-    description = models.TextField()
+    description = models.TextField(blank=True)
     item = models.FileField(upload_to='pages_files')
 
     def __str__(self):
--- a/wafer/pages/models.py
+++ b/wafer/pages/models.py
@@ -49,7 +49,8 @@ class Page(models.Model):
         help_text=_("Images and other files for use in"
                     " the content markdown field."))
 
-    people = models.ManyToManyField(settings.AUTH_USER_MODEL,
+    people = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
         related_name='pages', blank=True,
         help_text=_("People associated with this page for display in the"
                     " schedule (Session chairs, panelists, etc.)"))
--- a/wafer/pages/models.py
+++ b/wafer/pages/models.py
@@ -22,7 +22,11 @@ class File(models.Model):
     item = models.FileField(upload_to='pages_files')
 
     def __str__(self):
-        return u'%s' % (self.name,)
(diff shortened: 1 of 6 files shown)


wafer/pages/models.py
```diff
@@ -22,7 +22,11 @@ class File(models.Model):
     item = models.FileField(upload_to='pages_files')
 
     def __str__(self):
-        return u'%s' % (self.name,)
+        if self.pages.exists():
```
hodgestar: I'm wondering how sane it is to have `__str__` make database calls and whether this should rather be a separate method that is then explicitly specified in the admin interface?

wafer/sponsors/templatetags/sponsors.py
```diff
@@ -11,3 +11,17 @@ def sponsors():
...
     }
+
+
+# We use assignment_tag for compatibility with Django 1.8
+# Once we drop 1.8 support, we should change this to
+# simple_tag
+@register.assignment_tag(takes_context=True)
+def sponsor_image_url(context, name):
+    """Returns the corresponding url from the sponsors images"""
+    sponsor = context['sponsor']
```
hodgestar: Should we make the context explicit? E.g. `{% sponsor_image_url sponsor, "logo.png" as logo_url %}`?

wafer/pages/models.py
drnlm: The major issue is the display of the files in the sponsor admin interface, which uses a simple list. It isn't trivial to replace the use of str there with something else.

We could reduce this to just using the name and url, which should allow distinguishing files well enough.

wafer/sponsors/templates/wafer.sponsors/sponsors_block.html
```diff
@@ -6,8 +7,9 @@
             <ul>
                 {% for sponsor in package.sponsors.all %}
                     <li>
-                        {% if sponsor.logo %}
-                            <img src="{{ sponsor.logo.url }}" alt="{{ sponsor.name }}">
+                        {% sponsor_image_url sponsor "logo.png" as logo_url %}
```
stefanor: I don't think it's reasonable to assume it's going to be a png. I'm trying to use SVG where possible.

Shall we just call it `logo` ?

stefanor:
The annoyance of having to give the file a description is still present. Can we ditch the descriptions?

wafer/pages/models.py
hodgestar: I'm +1 on just using the name and URL.

hodgestar:
I'm +1 on ditching the descriptions if they annoy people. Maybe that's something for a separate PR though?

PR #336, opened 2017-01-25
Update years in the generated docs. Also change credit to 'wafer developement team', rather than just me

drnlm:
Our docs copyright info is somewhat outdated, so this updates it.


hodgestar:
Builds failed but changes look good. Build failures seem to all be for Django master, so I assume a change there broke something. Happy for this to land.

PR #338, opened 2017-01-26
Add sponsor name as a title of their logo (in case it's too tiny to read)

stefanor:
We did this for DC17, and it seems no-brainer enough...


drnlm:
Looks sensbile to me.

PR #340, opened 2017-01-27
Support for tracks

stefanor:
DebConf 17 would like to have some basic support for tracks (multiple talk types in multiple tracks results in a bit of a combinatorial explosion in the talk type drop-down).

The diff got a bit silly, because I tidied up some templates, so I'd suggest reviewing by commit or with [?w=1](https://github.com/CTPUG/wafer/pull/340/files?w=1)


hodgestar:
The changes look good. What happens to the talk submission form when a conference doesn't have an tracks?

stefanor:
> What happens to the talk submission form when a conference doesn't have an tracks?

Previously: an empty drop-down list.

Now: the field is hidden.

I'd be tempted to do this for talk types, too. They are rather a confusing stumbling block to enable talk submission.

drnlm:
I'm :+1: on doing the same for talk types.

stefanor:
Further IRC conversation:

```
18:08 <+tumbleweed> drnlm: that does mean making talk types optional, currently they aren't
19:47 <+drnlm> I think that's fine - not every conference has multiple talk types, and 
               requiring they set one may be an annoyance
19:48 < Hodgestar> Track chairs and session chairs and talk reviewers becomes very conference 
                   specific.
19:49 < Hodgestar> +1 on making talk types optional if there are no talk types set.
19:49 < Hodgestar> Should tracks and talks be required if they exist? (not at the db level 
                   but just at the submission form level?)
19:49 < Hodgestar> Organizers could add a "other" or "unknown" type option if they wanted?
19:50 <+drnlm> I think making them required if present makes the most sense
19:50 <+tumbleweed> you can have tracks, and still have talks that don't fit in tracks
19:50 <+tumbleweed> less so for types
19:51 <+tumbleweed> e.g. plenaries often don't fit in tracks
19:52 <+drnlm> But if the conference is accepting that, they can add "No track" or "Other" to 
               the list
19:53 <+drnlm> it's something we should document and describe how to deal with that case, but 
               I think must conference will want submitters to make a decision on which track 
               to submit to
19:53 <+drnlm> s/must/most/
19:54 <+tumbleweed> yeah
19:55 < Hodgestar> +1
19:57 <+drnlm> We probably should add templates that list the talk & track names and 
               descriptions - currently we ask for descriptions, but don't use them anywhere 
               by default
```

hodgestar:
Looks good to me!

drnlm:
Looks good to me

PR #343, opened 2017-01-30
Error page templates

### LINKED ISSUE #272: Wafer could implement 404.html and 500.html template pages
The problem with DebConf16. an apps had such templates, so it take precedence,

In any case, it would be nice to have a nice error page, with full links (menu).

lelutin:
The provides error pages with all of the basic wafer contents.

It aims to resolve #272 .

The templates are super simplistic, so make a sign if I need to add some more content to them.

### DIFF
--- a/wafer/templates/404.html
+++ b/wafer/templates/404.html
@@ -1,5 +1,8 @@
 {% extends "wafer/base.html" %}
+{% load i18n %}
 {% block content %}
-<h1>Page not found</h1>
+<h1>{% trans "Page not found" %}</h1>
+{% blocktrans %}
 The page you requested cannot be found.
+{% endblocktrans %}
 {% endblock %}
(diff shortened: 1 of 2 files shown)


wafer/templates/404.html
```diff
@@ -0,0 +1,5 @@
+{% extends "wafer/base.html" %}
+{% block content %}
+<h1>Page not found</h1>
+The page you requested cannot be found.
```
stefanor: Can you add i18n, please? (trans/blocktrans)

We try to tag everything that's user-visible for translation.

lelutin: oh indeed! that's a great comment. I'll add this later today

lelutin:
@stefanor I've pushed a commit on top of my branch that adds translate functions to the template. I've used blocktrans in the text "body" so that it's easier to extend if we think of something longer and more descriptive to write there.

PR #344, opened 2017-01-31
Describe our approach to authors

stefanor:
Mention that the submitter will be the corresponding author.
And that explain the privacy mode, if enabled.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -31,8 +31,9 @@ def authors_help():
     text.append(_("The speakers presenting the talk."))
     if not settings.WAFER_PUBLIC_ATTENDEE_LIST:
         text.append(_(
-            "To ensure privacy, you will only be able to see yourself and "
-            "authors that have been added by the conference organisers. "
+            "To ensure attendee privacy, you will only be able to see "
+            "yourself and authors that have been added to the talk by the "
+            "conference organisers. "
             "If you will have other co-authors, add a note in the notes "
             "field, so the organisers can add them to your talk."
         ))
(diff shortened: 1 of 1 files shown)


wafer/talks/models.py
```diff
@@ -24,6 +25,23 @@ def render_author(author):
...
 
 
+def authors_help():
+    _ = ugettext  # This function will be wrapped for lazy evaluation
+    text = []
+    text.append(_("The speakers presenting the talk."))
+    if not settings.WAFER_PUBLIC_ATTENDEE_LIST:
+        text.append(_(
+            "To ensure privacy, you will only be able to see yourself and "
+            "authors that have been added by the conference organisers. "
```
drnlm: I think it's worth emphasising that this is a "per talk" thing, It's not completely clear from the text, and people with multiple submission may be suprised that the list of available co-authors differs between talks.

wafer/talks/models.py
```diff
@@ -24,6 +25,24 @@ def render_author(author):
...
+    if not settings.WAFER_PUBLIC_ATTENDEE_LIST:
+        text.append(_(
+            "To ensure attendee privacy, you will only be able to see "
+            "yourself and authors that have been added to the talk by the "
+            "conference organisers. "
+            "If you will have other co-authors, add a note in the notes "
+            "field, so the organisers can add them to your talk."
+        ))
+    text.append(_(
+        "You, as the talk submitter, will be the talk's corresponding author."
```
drnlm: With the full text displayed, this note gets a bit lost at the end of it. I think it should be emphasised in some way - perhaps wrapped in a <strong> tag or similar?

PR #346, opened 2017-01-31
Add video release permission (and reviewer) to talks

stefanor:
We really should ask our speakers for approval for recording video of their talks. This lets that be explicit.

Relatedly, @CarlFK has been thinking about how to get people to verify the information in abstracts, so that the video title slides come out well. He is [suggesting](http://whiteboard.debian.net/fdda53.wb) the addition of a video reviewer field.

He proposes having a video reviewer, independent from the talk submitter. Here's how he describes this:

> *Reviewer* or *Personal QA*
> 
> help_text: email of a person who will be allowed to review and approve your video details.
>
> A person who will be allowed to review and approve the video on behalf of the presenter.  A second set of eyes who is not burdened with preparing to give a talk, travelling to and attending a conference, talking to people in the hallway and all the other things a presenter does.  This person will be one of the first people to see the final video as soon as it is uploaded to YouTube.   It will shortcut the current process of waiting for someone from the on site video crew doing a quick review, which sometimes doesn't happen for hours or days later.   This person will never be a blocker, but will be given the ability to click the Approve button for the one video they have been chosen for.  They will also be CCed the "please review your title slide" email shortly before the event.  The assumption is they are aware of the topic and will be able to spot typos or incorrect punctuation of the project name and presenter's name, or odd word wrapping any of the other handful of things people might care about.   
>
> The primary focus of this role is to assist in getting the video made public sooner.    The on site video crew has plenty to do setting up and recording the talks.  If everything goes perfectly, we wouldn't need to double check anything.  But in reality there are problems that need to be caught, and the process of catching them is time consuming, and has to be applied to every single item, not just the ones with problems.  This person will also be the one to notice a video has fallen though the cracks and forgotten about.     
> 
> So to help increase the quality and timeliness, please select a person.  Ideally it is someone not planning on attending the event and is eagerly awaiting to see this video.  If they ignore everything, that's OK, things will be processed as before.  
> 
> Note this is a 1:1 mapping.  We should not have the same person assigned to more than one event.  We want all of their effort going into this one item, not divided across two. 
>
> ### for conference organizers
> 
> Note this applies to Keynotes, Lightning talks (either if they are separate or are lumped together into one item), conference opening and closing, BOFs, meetings and anything else that should become a video.   It will help define what should and should not be recorded and produced.  These edge cases are the most likely candidates for problems, so the most in need of someone to look after them.  If there is an event but no one can be found to fill this role, it brings into question if we should bother producing a video.

That's way too long to put in a help text, it really should be described in a stand alone web page somewhere, that can be linked to. I tried my best, to summarise it down to a couple of sentences for the help text field.


drnlm:
Adding a explicit question about video'ing talks seems fine.

I'm not sold on tying this feature to also adding the video reviewer field. This feels tightly coupled to the way we've handled vidoes at the last couple of PyCon ZA's, rather than being generically useful. Other approaches, such using a professional video team who will review the video, or having the conference organisers review all the videos, don't fit particularly well with this as it stands.

stefanor:
> I'm not sold on tying this feature to also adding the video reviewer field.

Throw it behind another config flag?

This could be something a conference implemented via an extra table or KV. But that's a fair bit of work and would probably require a lot of duplication.

drnlm:
> Throw it behind another config flag?

I worry about the complexity that introduces - we do want to keep the number of options people need to fiddle with sane. I think we need to let the reviewer idea bake a bit more before committing to an implementation, especially as we'll probably need to extend the solution to include pages to cater for videos of special events.

drnlm:
I can live with it behind a config flag.

PR #347, opened 2017-02-07
The Debian SSO URL changed

stefanor:


stefanor:
Indeed. It would have been useful if there was a redirect.

Basically, this used to be a "beta" new way of doing SSO, and it got promoted.

PR #352, opened 2017-04-12
Replace KV registration form with custom registration API

stefanor:
I wasn't very satisfied with the KV registration approach we used for DebConf16. While this framework made implementing the form really nice and easy, the data was completely unusable in Django admin, which made dealing with the data a lot harder.

I think if you find yourself needing some sort of registration form, you're really better off throwing a django model + form together, yourself. So, here's a simple API to let wafer know if a user is registered, when you have done that.

We could leave the KV registration framework in place, but I don't know if I want to inflict it on anyone else :)


wafer/users/models.py
```diff
@@ -22,6 +23,7 @@
 # Specification taken from https://support.twitter.com/articles/101299
 TwitterValidator = RegexValidator('^[A-Za-z0-9_]{1,15}$',
                                   'Incorrectly formatted twitter handle')
+is_registered = import_string(settings.WAFER_USER_IS_REGISTERED)
```
drnlm: Is it worth catching ImportError and returning a message that explicitly points to the WAFER_USER_IS_REGISTERED setting, rather than just the raw error?

wafer/settings.py
```diff
@@ -280,11 +280,12 @@
...
-# Can be 'ticket' for Quicket tickets or 'form' for a classic form
-WAFER_REGISTRATION_MODE = 'ticket'
 
-# For REGISTRATION_MODE == 'form', the form to present
-WAFER_REGISTRATION_FORM = 'wafer.users.forms.ExampleRegistrationForm'
+# WAFER_REGISTRATION_MODE can be 'ticket' for Quicket tickets, or 'custom' if
+# you implement your own registration system.
+WAFER_REGISTRATION_MODE = 'ticket'
+# WAFER_USER_IS_REGISTERED should return a boolean, when passed a Django user.
+WAFER_USER_IS_REGISTERED = 'wafer.tickets.models.user_is_registered'
```
drnlm: We should add this to the documentation, rather than just having comments in the settings file

hodgestar:
@stefanor Should we try land this?

PR #359, opened 2017-07-24
Hacky approach to caching the page rendering for dynamic page content

hodgestar:
The recent addition of support for dynamic variables to markdown in PyConZA means that the approach of caching rendered markdown might need to be revisited.

There are a few options and we'll have to trade off code complexity against supporting dynamic or static markdown pages.

Possibly we want to allow a cache age to be set on pages so that we know when to invalidate cached pages and can continue to cache static markdown forever by default.


drnlm:
My vague plan is to add an extra field to the Page model to control the cache timeout, with some marker value to retain the current behaviour.

stefanor:
Have you seen [cacheback](https://github.com/codeinthehole/django-cacheback)? It's a neat way to avoid thundering herds when the page expires.

drnlm:
While cacheback looks nice, I'm a bit worried that adding support for it is a fairly heavy dependency for wafer.

wafer/pages/models.py
```diff
@@ -68,6 +78,24 @@ def get_absolute_url(self):
         url = "/".join(self.get_path())
         return reverse('wafer_page', args=(url,))
 
+    def cached_render(self):
+        if self.cache_time < 0:
+            return self.content.rendered
+        cache = caches[self.cache_name]
+        cache_key = self.get_absolute_url()
```
hodgestar: If multiple parts of wafer are using the cache for different things, should we be prefixing the key names with some sort of namespace string?

wafer/pages/models.py
```diff
@@ -6,11 +6,12 @@
 from django.core.exceptions import ValidationError, NON_FIELD_ERRORS
 from django.conf import settings
 from django.db import models
+from django.core.cache import caches
 from django.db.models.signals import post_save
 from django.utils.encoding import python_2_unicode_compatible
 
 
-from markitup.fields import MarkupField
+from markitup.fields import MarkupField, render_func
```
hodgestar: Just wanted to double-check that render_func gets all the settings and extensions from settings since we're not actually testing that and the initial rendering goes via the MarkupField base logic.

wafer/pages/models.py
drnlm: That's a good point, and we should probably do that.

Do you think 'page_' is sufficient?

wafer/pages/models.py
drnlm: django-markitup creates render_func with all the settings (see https://github.com/zsiciarz/django-markitup/blob/master/markitup/fields.py), and then uses it for the default rendering throughout.

We're copying the logic from MarkupField's pre_save method here, so we should be fine.

wafer/pages/models.py
hodgestar: Maybe something like ``wafer.pages:rendered:...``? Just thinking that the module name is a nice existing namespace and `:` isn't a valid part of a module name.

wafer/pages/models.py
hodgestar: Woot.

hodgestar:
@stefanor Re cacheback -- I think it would be cool to add optional support for cacheback but I think that can be done on top of the work in this PR without having to change the DB model again?

hodgestar:
:+1: (I can't use GitHub's approve button because I opened the PR).

Issue #358, opened 2017-07-19
Talks scheduled before the speaker arrives to the conference should show an error


baldurmen:
While working on the DebConf17 schedule, someone had to go talk by talk and verify manually if the talk was scheduled before the speaker actually arrived to the conference.

I think modifying the custom registration API showcased in #352 to add an "arriving date" and a "departure date" and linking this to the schedule validation would be highly beneficial and time saving.

Maybe having a "confirmed attendance" API (we are using this for DebConf to validate that people will actually be there) would be interesting to validate too?

stefanor:
> Maybe having a "confirmed attendance" API (we are using this for DebConf to validate that people will actually be there) would be interesting to validate too?

I think that'll be resolved in #352, when I get around to finishing that.

drnlm:
At some point, features start becoming very conference-specific, and it's not sensible to try directly support them in wafer itself. Given that this will require using a custom registration profile, I feel that this is such a feature, and Debconf should be extending the schedule validation logic locally to support this specific request.

wafer should make it easy to extend or otherwise tweak the schedule validation logic, though, so suggestions on how to improve that welcome.

hodgestar:
Maybe we could add some sort of custom schedule validation hook?

PR #361, opened 2017-07-27
Feature/custom schedule validators

drnlm:
To help address #358 , this refactors to schedule validation code to use explicit lists of validation functions.

This should make it much simpler to add custom validators to extend the behaviour of the schedule.

### DIFF
--- a/docs/schedule.rst
+++ b/docs/schedule.rst
@@ -82,7 +82,7 @@ schedule item validator is called with a list of all schedule items.
 Validators are expected to return a list of invalid items or an
 empty list if the validator finds no error.
 
-Use ``register_schedule_item_validator`` and ``register_slot_vaildator``
+Use ``register_schedule_item_validator`` and ``register_slot_validator``
 to add the validators to the list.
 
 To display the errors in the admin form, you will also need to extend the
--- a/docs/schedule.rst
+++ b/docs/schedule.rst
@@ -69,3 +69,21 @@ Schedule items which are not talks have ``talk-type-none`` as the CSS class.
 
 A per item CSS class can also be set using the ``css_class`` attribute on the
 schedule item.
+
+
+Adding additional schedule validation
+=====================================
+
+Wafer runs validation on the slots and the schedule items. This behaviour
+can be extended by providing custom validators.
+
+Each validator is called with a list of all the items and is expected to
+return a list of invalid items or an empty list if the validator finds
+no error.
+
+Use ``register_schedule_item_validator`` and ``register_slot_vaildator``
+to add the validators to the list.
+
+To display the errors in the admin form, you will also need to extend the
+``displayerrors`` block in ``scheduleitem_list.html`` and ``slot_list.html``
+templates.
--- a/docs/schedule.rst
+++ b/docs/schedule.rst
@@ -77,9 +77,10 @@ Adding additional schedule validation
 Wafer runs validation on the slots and the schedule items. This behaviour
 can be extended by providing custom validators.
 
-Each validator is called with a list of all the items and is expected to
-return a list of invalid items or an empty list if the validator finds
-no error.
+Each slot validator is called with a list of all the slots, and each
+schedule item validator is called with a list of all schedule items.
+Validators are expected to return a list of invalid items or an
+empty list if the validator finds no error.
 
 Use ``register_schedule_item_validator`` and ``register_slot_vaildator``
 to add the validators to the list.
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -147,25 +147,40 @@ def prefetch_slots():
     return list(Slot.objects.all())
 
 
+# Validators are listed as (function, error type, error message) tuples
+SLOT_VALIDATORS = [
+   (find_overlapping_slots, 'overlaps',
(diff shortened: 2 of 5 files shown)


docs/schedule.rst
```diff
@@ -69,3 +69,21 @@ Schedule items which are not talks have ``talk-type-none`` as the CSS class.
...
+=====================================
+
+Wafer runs validation on the slots and the schedule items. This behaviour
+can be extended by providing custom validators.
+
+Each validator is called with a list of all the items and is expected to
+return a list of invalid items or an empty list if the validator finds
+no error.
+
+Use ``register_schedule_item_validator`` and ``register_slot_vaildator``
```
hodgestar: s/vail/vali/

docs/schedule.rst
```diff
@@ -69,3 +69,21 @@ Schedule items which are not talks have ``talk-type-none`` as the CSS class.
...
 schedule item.
+
+
+Adding additional schedule validation
+=====================================
+
+Wafer runs validation on the slots and the schedule items. This behaviour
+can be extended by providing custom validators.
+
+Each validator is called with a list of all the items and is expected to
```
hodgestar: Should we explain that item validators get a list of items and slot validators get a list of slots?

wafer/schedule/admin.py
```diff
@@ -151,41 +143,76 @@ def prefetch_schedule_items():
...
+
+
+# Validators are listed as (function, error type, error message) tuples
+SLOT_VALIDATORS = []
+SCHEDULE_ITEM_VALIDATORS = []
+
+
+# Helpers for people extending the tests
+def register_slot_validator(function, err_type, msg):
+    global SLOT_VALIDATORS
```
hodgestar: Given that we're calling `.append(...)` we don't actually need the `global`s here but I guess it's not a bad idea to make it clear that these are globals. Happy either way, just checking intent.

wafer/schedule/admin.py
```diff
@@ -218,23 +245,12 @@ class ScheduleItemAdmin(admin.ModelAdmin):
...
-        if validation:
-            errors['validation'] = validation
-        if venues:
-            errors['venues'] = venues
-        if non_contiguous:
-            errors['non_contiguous'] = non_contiguous
+        for validator, err_type, _msg in SCHEDULE_ITEM_VALIDATORS:
+            failed_items = validator(all_items)
+            if failed_items:
+                errors[err_type] = failed_items
```
hodgestar: I'm a little worried this can override errors if `err_type` is not unique (and the registration interface doesn't prevent that). Maybe we could do:
```python
errors = defaultdict(list)
...
errors[err_type].extend(failed_items)
```
?

wafer/schedule/admin.py
```diff
@@ -319,9 +335,11 @@ def changelist_view(self, request, extra_context=None):
...
         # Find issues with the slots
         errors = {}
-        overlaps = find_overlapping_slots()
-        if overlaps:
-            errors['overlaps'] = overlaps
+        all_slots = prefetch_slots()
+        for validator, err_type, _msg in SLOT_VALIDATORS:
+            failed_slots = validator(all_slots)
+            if failed_slots:
+                errors[err_type] = failed_slots
```
hodgestar: Same question about uniqueness of `err_type` here.

wafer/schedule/templates/admin/slot_list.html
```diff
@@ -8,6 +8,7 @@
        {% if errors %}
        <div name="errors">
           <h2>{% trans "Errors in the slots" %}</h2>
+          {% block displayerrors %}
           {% if errors.overlaps %}
           <h3>{% trans "OVerlapping slots" %}</h3>
```
hodgestar: s/OVer/Over/

docs/schedule.rst
```diff
@@ -69,3 +69,21 @@ Schedule items which are not talks have ``talk-type-none`` as the CSS class.
...
+return a list of invalid items or an empty list if the validator finds
+no error.
+
+Use ``register_schedule_item_validator`` and ``register_slot_vaildator``
+to add the validators to the list.
+
+To display the errors in the admin form, you will also need to extend the
```
hodgestar: Not necessarily for this PR, but it would be cool if the default `displayerrors` block at least reported all errors so that people didn't need to override this block just to see their custom errors.

wafer/schedule/admin.py
drnlm: I personally prefer the explicit globals declaration for clarity.

wafer/schedule/admin.py
drnlm: Good point - I'll make that change.

docs/schedule.rst
drnlm: It would be, but there's some custom rendering behaviour for clashes which makes that a bit tricky.

wafer/schedule/admin.py
hodgestar: Cool.

docs/schedule.rst
hodgestar: I'm guessing we could keep the special casing but still display everything else, but let's leave that for a later PR.

PR #368, opened 2017-08-11
Add a published flag

stefanor:
So pages can be staged in git. While we're here, document the flags.

### DIFF
--- a/docs/pages.rst
+++ b/docs/pages.rst
@@ -68,7 +68,8 @@ e.g. ``pages/index.md``::
 The front matter can contain a couple of flags:
 
 ``published``
-    If set to ``false``, the page will not be loaded.
+    If set to ``false``, the page will not be loaded by the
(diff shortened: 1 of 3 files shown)


docs/pages.rst
```diff
@@ -64,3 +64,13 @@ e.g. ``pages/index.md``::
 
     We invite you to [join us](/attend/) at [our venue](/venue/)
     on the 31st of December for a day of fun conferencing.
+
+The front matter can contain a couple of flags:
+
+``published``
+    If set to ``false``, the page will not be loaded.
```
hodgestar: Is `loaded` the right verb here? I'm not quite following what the sentence is intended to convey.

hodgestar: Oh, I see. Perhaps we should explain that it is not loaded by the `load_pages` management command?

PR #371, opened 2017-08-23
Add suitably marked section around speakers to allow styling the page.

hodgestar:


hodgestar:
I've confirmed that the section element doesn't mess with the layout overly. Would appreciate feedback on whether the divs around the logo and the name are the right ways to make those stylable.

PR #372, opened 2017-08-23
Explicitly set on_delete to models.CASCADE for ForeignKeys.

hodgestar:
As described in https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete explicitly setting `on_delete` is required in Django 2.0.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -112,8 +112,10 @@ class Talk(models.Model):
     )
 
     talk_id = models.AutoField(primary_key=True)
-    talk_type = models.ForeignKey(TalkType, null=True, blank=True)
-    track = models.ForeignKey(Track, null=True, blank=True)
+    talk_type = models.ForeignKey(
+        TalkType, null=True, blank=True, on_delete=models.CASCADE)
+    track = models.ForeignKey(
+        Track, null=True, blank=True, on_delete=models.CASCADE)
 
     title = models.CharField(max_length=1024)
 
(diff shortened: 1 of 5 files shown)


wafer/talks/models.py
```diff
@@ -112,8 +112,10 @@ class Meta:
     )
 
     talk_id = models.AutoField(primary_key=True)
-    talk_type = models.ForeignKey(TalkType, null=True, blank=True)
-    track = models.ForeignKey(Track, null=True, blank=True)
+    talk_type = models.ForeignKey(
+        TalkType, null=True, blank=True, on_delete=models.CASCADE)
+    track = models.ForeignKey(
+        Track, null=True, blank=True, on_delete=models.CASCADE)
```
stefanor: Maybe these should be PROTECT or SET_NULL?

hodgestar: I was thinking the same but thought I should initially set them to exactly what they are now. Happy to change them to `SET_NULL` since they're allowed to be null currently anyway?

stefanor: Yeah, I think that's the reasonable thing to do here.

hodgestar:
@stefanor Are you happy for me to land these without migrations? There are some outstanding migrations but those include migrations that should have happened in other branches and maybe some that are just Django shifting.

Perhaps we need a separate migrations clean-up PR?

PR #375, opened 2017-09-26
Tweak logic around clashes and invalid venues, so errors show correctly in the admin interface

drnlm:
The current logic for displaying errors loses the details for clashes and invalid venues, which makes debugging schedule issues harder than it should be.

This fixes the missing information problem.


drnlm:
The changes to the formatting are so the output better matches the list in the admin interface, which makes searching for the cause of the error simpler.

hodgestar:
Except that tests failed.

drnlm:
Tests fixed

PR #376, opened 2017-09-26
Mark cancelled talks in the Schedule editor.

drnlm:
While we may want to manipulate cancelled talks using the schedule
editor, they should be clearly marked as such.

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/edit_schedule.html
+++ b/wafer/schedule/templates/wafer.schedule/edit_schedule.html
@@ -131,7 +131,7 @@
                   data-toggle="tooltip" data-placement="left"
                   title="{{ talk.title }} (Cancelled)" data-type="talk" data-talk-id="{{ talk.talk_id }}">
               {% if not talk.cancelled %}
-              {{ talk.title|truncatechars:24 }} Isn't cancelled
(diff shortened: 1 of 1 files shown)


wafer/schedule/templates/wafer.schedule/edit_schedule.html
```diff
@@ -121,8 +129,12 @@
             <span draggable="true" class="col-md-6 badge badge-warning draggable"
                   id="talk{{ talk.talk_id }}"
                   data-toggle="tooltip" data-placement="left"
-                  title="{{ talk.title }}" data-type="talk" data-talk-id="{{ talk.talk_id }}">
-              {{ talk.title|truncatechars:24 }}
+                  title="{{ talk.title }} (Cancelled)" data-type="talk" data-talk-id="{{ talk.talk_id }}">
+              {% if not talk.cancelled %}
+              {{ talk.title|truncatechars:24 }} Isn't cancelled
```
hodgestar: I'm not following why we need the "Isn't cancelled" here?

drnlm: That was left over debugging output - removed.

PR #381, opened 2017-10-08
Document settings

stefanor:
Prerequisite for dealing with the feedback in #352.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -73,7 +73,8 @@ Wafer's settings
 ``WAFER_PUBLIC_ATTENDEE_LIST``
     A boolean flag.
     When ``True``, all registered users' profiles are publicly visible.
-    Otherwise, only users with associated talks have public profiles.
+    Otherwise, only users with associated public talks have public
+    profiles.
 
 ``WAFER_REGISTRATION_FORM``
     A Django Form used for KV registration.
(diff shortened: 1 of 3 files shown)


docs/settings.rst
```diff
@@ -0,0 +1,142 @@
...
+        The URL to link to.
+
+    ``items``
+        An optional list of similar dicts, making up a submenu.
+
+``WAFER_PUBLIC_ATTENDEE_LIST``
+    A boolean flag.
+    When ``True``, all registered users' profiles are publicly visible.
+    Otherwise, only users with associated talks have public profiles.
+
```
drnlm: "associated public talks", perhaps - we don't expose users until we've accepted their talk.

docs/settings.rst
```diff
@@ -0,0 +1,142 @@
...
+``PAGE_DIR``
+    The directory that the ``load_pages`` management command will load
+    pages from.
+    Should be an absolute path with a trailing ``/``.
+
+``TALKS_OPEN``
+    Controls whether talk submissions are accepted.
+    Set to ``False`` to close talk submissions.
+
```
drnlm: Do we want to mention that there's fine grained control avaible for individual talk types via the admin interface? Not sure if this is the best place to mention that, though

stefanor: I think we should have a page on conference paper management.

PR #391, opened 2017-12-17
Feature/add ical export

drnlm:
This adds a simplistic ical export of the schedule.


wafer/schedule/admin.py
```diff
@@ -221,6 +221,7 @@ def validate_schedule():
 class ScheduleItemAdminForm(forms.ModelForm):
     class Meta:
         model = ScheduleItem
+        readonly_fields = ('last_updated')
```
hodgestar: Is this missing a trailing comma?

wafer/schedule/models.py
```diff
@@ -261,6 +281,25 @@ def invalidate_check_schedule(*args, **kw):
...
+    for item in slot.scheduleitem_set.all():
+       item.save()
+    # We also need to update the next slot, in case we changed it's
+    # times as well
+    next_slot = slot.slot_set.all()
+    if next_slot.count():
+        # From the way we structure the slot tree, we know that
+        # there's only 1 next slot that could have changed.
+        for item in next_slot[0].scheduleitem_set.all():
+            item.save()
```
hodgestar: Are we sure that we can't trigger a save loop if two slots form a cycle?

wafer/schedule/admin.py
drnlm: Yes, it is

wafer/schedule/models.py
drnlm: The validation on slots should ensure we can't have a slot cycle, but restricting these to only save the update_date is probably a good idea.

drnlm:
Comments hopefully addressed.

PR #395, opened 2017-12-28
Towards Django 2.0 support 

stefanor:
Mostly adding `on_delete` to migrations. Also, follow through on some other deprecations.

This seems to work, at least for basic functionality, haven't tried building a schedule.

`drf-extensions` doesn't support Django 2.0 yet, I tested with local hacks to it. I see the same things in master, so guess it should at least be importable in the next release (although possibly not working correctly, yet).
`django-markitup` doesn't support Django 2.0 yet, but I did some work on that, upstream: zsiciarz/django-markitup#21.

### DIFF
--- a/wafer/talks/migrations/0001_initial.py
+++ b/wafer/talks/migrations/0001_initial.py
@@ -16,14 +16,30 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='Talk',
             fields=[
-                ('talk_id', models.AutoField(serialize=False, primary_key=True)),
+                ('talk_id', models.AutoField(
+                    serialize=False, primary_key=True)),
                 ('title', models.CharField(max_length=1024)),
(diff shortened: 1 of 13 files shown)


wafer/talks/migrations/0001_initial.py
```diff
@@ -23,7 +23,7 @@ class Migration(migrations.Migration):
                 ('status', models.CharField(default=b'P', max_length=1, choices=[(b'A', b'Accepted'), (b'R', b'Not Accepted'), (b'P', b'Under Consideration')])),
                 ('_abstract_rendered', models.TextField(editable=False, blank=True)),
                 ('authors', models.ManyToManyField(related_name='talks', to=settings.AUTH_USER_MODEL)),
-                ('corresponding_author', models.ForeignKey(related_name='contact_talks', to=settings.AUTH_USER_MODEL)),
+                ('corresponding_author', models.ForeignKey(related_name='contact_talks', to=settings.AUTH_USER_MODEL, on_delete=models.CASCADE)),
```
drnlm: Any specific reason to not reformat this long line like the other changes?

stefanor: I think I meant to come back to that. The whole area is a lot messier.

PR #396, opened 2018-01-30
Add KeyValue.__str__ for python 3

stefanor:


drnlm:
why aren't we using python_2_unicode_compatible (as we do elsewhere) here?

PR #397, opened 2018-01-30
Improve the talk submission form

nbraud:
Based on feedback on the wafer install for DebConf 18.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -129,7 +129,8 @@ class Talk(models.Model):
 
     notes = models.TextField(
         null=True, blank=True,
-        help_text=_("Any notes for the conference organisers?"))
+        help_text=_("Any notes for the conference organisers?  "
+                    "These are not visible to the public."))
 
     private_notes = models.TextField(
         null=True, blank=True,


wafer/talks/models.py
```diff
@@ -129,7 +129,8 @@ class Meta:
 
     notes = models.TextField(
         null=True, blank=True,
-        help_text=_("Any notes for the conference organisers?"))
+        help_text=_("Any notes for the conference organisers?  "
+                    "Attendees cannot read those."))
```
stefanor: The "those" seems a bit distant. How about: These are not visible to attendees or the public.

wafer/static/css/wafer.css
```diff
@@ -47,6 +47,10 @@
     border: 5px dashed #000;
 }
 
+.help-block {
+    color: #d9534f;
```
stefanor: That's quite red. Is that not going to be too close to the colour of errors?

nbraud: I picked it because it was the color of errors: `.help-block` is set on the validation error messages.

nbraud: Yeap, confirmed (and removed it from the patch).

stefanor: err it still seems to be here.

wafer/talks/models.py
```diff
@@ -129,7 +129,8 @@ class Meta:
 
     notes = models.TextField(
         null=True, blank=True,
-        help_text=_("Any notes for the conference organisers?"))
+        help_text=_("Any notes for the conference organisers?  "
+                    "Those are not visible to the public."))
```
drnlm: I don't like the use of "Those" here - I would prefer "These are not visible ..."

nbraud: Fixed.

PR #399, opened 2018-01-31
Load jQuery once and only once, in page foot

stefanor:
I found a small herd of hairy yaks while looking at a JS error on our talk submission page:

> Uncaught TypeError: element.markItUp is not a function

The form was rendering inline script and CSS tags, that depended on JS that we were only loading at the end of the page.

Fixing some of this is fairly simple:
1. Pages with forms get slightly more complex, and are instructed to render CSS in the HEAD, and JS in the page foot.
1. Widgets are instructed to not load their own jQuery, because we provide one.

However, markitup renders inline JS that requires jQuery and markitup's JS library that defeats this strategy (zsiciarz/django-markitup#25) so I implemented a template override to work around it. I'll propose the same changes upstream.

### DIFF
--- a/wafer/templates/markitup/editor.html
+++ b/wafer/templates/markitup/editor.html
@@ -1,0 +1,4 @@
+<div class="django-markitup-config" style="display: none"
+     data-element="#{{ textarea_id }}"
+     data-preview-url="{{ preview_url }}"
+     data-auto-preview="{{ AUTO_PREVIEW|yesno:"1,0" }}"></div>
--- a/wafer/compare/templates/admin/wafer.compare/change_form.html
+++ b/wafer/compare/templates/admin/wafer.compare/change_form.html
@@ -1,5 +1,6 @@
 {% extends "admin/change_form.html" %}
 {% load i18n admin_urls %}
+{% load static from staticfiles %}
 {% block object-tools-items %}
     <li>
         <a href="{% url opts|admin_urlname:'history' original.pk|admin_urlquote %}" class="historylink">{% trans "History" %}</a>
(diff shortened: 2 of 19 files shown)


wafer/templates/markitup/editor.html
```diff
@@ -0,0 +1,4 @@
+<div class="django-markitup-config" style="display: none"
+     data-element="#{{ textarea_id }}"
+     data-preview-url="{{ preview_url }}"
+     data-auto-preview="{{ AUTO_PREVIEW|yesno:"1,0" }}"></div>
```
drnlm: How does this change interact with Markitup's Admin widget?

Won't we need to also ensure the javascript snippet is included on the admin pages so this works there too?

stefanor: Good question. I wanted to test that too :)

stefanor: Resolved that by hooking into `wafer/compare/templates/admin/wafer.compare/change_form.html`. But that's a temporary hack if upstream will accept a similar change.

stefanor:
Upstreamed the markitup hacks as zsiciarz/django-markitup#26 (although upstream seems pretty dead these days)

wafer/templates/markitup/editor.html
drnlm: Playing with this branch, it looks like we aren't doing the right thing for the markup fields used in the sponsors admin page, but otherwise this looks good.

stefanor: Overrode that admin template too.

stefanor:
If this is ever fixed upstream, I propose merging https://github.com/CTPUG/wafer/compare/roll-back-markitup-hacks?expand=1

PR #402, opened 2018-02-01
The well is dry (more bootstrap4 migration)

stefanor:
There were a few leftover clasess from bootstrap 3 that we never stopped using after bootstrap 4 dropped them.

* [well][]s are largely replaced by [card][]s.
* `bs-callout` was something we copied from Bootstrap3 documentation, but it didn't really fit into cards.

[well]: https://getbootstrap.com/docs/3.3/components/#wells
[card]: https://getbootstrap.com/docs/4.0/components/card/

I've tried to tag objects with classes, so that consumers of wafer can style the elements to suit their site. There are probably more places that we need to do this...

Unfortunately, there's a lot of whitespace change involved. I tried to tidy indentation while I was there, but the diff is going to be hard to read.

### DIFF
--- a/wafer/talks/templates/wafer.talks/talk.html
+++ b/wafer/talks/templates/wafer.talks/talk.html
@@ -82,22 +82,34 @@
   </div>
   {% if perms.talks.view_all_talks or user.is_superuser %}
     {% if talk.notes %}
-      {% blocktrans %}
-        <h2>Talk Notes</h2>
-        <p>(The following is not visible to attendees.)</p>
-      {% endblocktrans %}
-      <div id="notes">
-        {{ object.notes|urlize|linebreaks }}
+      <div id="notes" class="card mb-3">
+        <div class="card-header">
+          {% blocktrans %}
+            <h2>Talk Notes</h2>
+            <p>(The following is not visible to attendees.)</p>
+          {% endblocktrans %}
+        </div>
+        <div class="card-body">
+          <p class="card-text">
+            {{ object.notes|urlize|linebreaksbr }}
+          </p>
+        </div>
       </div>
     {% endif %}
   {% endif %}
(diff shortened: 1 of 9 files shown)


wafer/users/templates/wafer.users/profile.html
```diff
@@ -1,165 +1,192 @@
...
+                 class="float-right btn btn-secondary btn-lg">
+                {% trans 'Edit' %}
+              </a>
+              <h3 class="card-title">
+                <a href="{{ talk.get_absolute_url }}">
+                  {{ talk.title }}
+                </a>
+              </h3>
+              <p>{{ talk.abstract.rendered|safe }}</p>
+            </div>
```
drnlm: We should revisit this template and see if we can't reduce the duplication between the different talk lists somehow.

Out of scope for this PR, though.

wafer/talks/templates/wafer.talks/talk.html
```diff
@@ -86,7 +86,7 @@ <h1>
         <h2>Talk Notes</h2>
         <p>(The following is not visible to attendees.)</p>
       {% endblocktrans %}
-      <div class="well">
+      <div id="notes">
```
drnlm: Shouldn't these sections also become cards, to better how things behaved with bootstrap 3?

stefanor: Yeah, I went through that thought process too. Cards didn't look right for the abstract, in the debconf17 theme, at the time.

stefanor: But for notes, they may make sense

wafer/users/templates/wafer.users/profile.html
stefanor: Yeah, that'd be nice.

PR #413, opened 2018-04-10
Use django-registration's backend URLs

stefanor:
django-registration has supported Django 1.5 for about 5 years now. There's no need to re-implement this ourselves.

Plus, we can get some new features like /accounts/activate/resend/


stefanor:
Those test failures should be fixed by #414

PR #415, opened 2018-04-10
Add templates to show talk types and tracks and their descriptions.

drnlm:
So we can link to them elsewhere, such as a "how to submit a talk" page or the like.

### DIFF
--- a/wafer/talks/templates/wafer.talks/talk_tracks.html
+++ b/wafer/talks/templates/wafer.talks/talk_tracks.html
@@ -4,11 +4,15 @@
 {% block content %}
 <section class="wafer wafer-talk-tracks">
   <h1>{% trans 'Tracks' %}</h1>
+    {% if object_list.count > 0 %}
     <dl class="track-list">
     {% for track in object_list %}
       <dt class="track-name">{{ track.name }}</dt>
       <dd class="track-description">{{ track.description }}</dt>
     {% endfor %}
     </dl>
+    {% else %}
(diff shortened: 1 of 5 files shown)


hodgestar:
Woot. Given that these aren't linked to from anywhere by default, should we mention them in the documentation somewhere so that people know they exist?

hodgestar:
Separate thought: Is it worthwhile adding a markup extension that allows generic access to something like Django's `url_for` in templates?

drnlm:
A markdown extension supporting url_for sounds like a great addition.

wafer/talks/templates/wafer.talks/talk_tracks.html
```diff
@@ -0,0 +1,15 @@
+{% extends "wafer/base.html" %}
+{% load i18n %}
+{% block title %}{% trans "Tracks" %} - {{ WAFER_CONFERENCE_NAME }}{% endblock %}
+{% block content %}
+<section class="wafer wafer-talk-tracks">
+  <h1>{% trans 'Tracks' %}</h1>
+    <ul class="track-list">
+    {% for track in object_list %}
+       <li>{{ track.name }} &mdash; {{ track.description }}</li>
```
stefanor: Did you consider using HTML definition lists?

drnlm: No - it's one of those features of HTML I always forget about.

PR #418, opened 2018-04-11
Talk reviews

stefanor:
Allow a content team to vote on talks.


wafer/talks/admin.py
```diff
@@ -48,7 +49,7 @@ class TalkUrlInline(admin.TabularInline):
 class TalkAdmin(CompareVersionAdmin, admin.ModelAdmin):
     list_display = ('title', 'get_corresponding_author_name',
                     'get_corresponding_author_contact', 'talk_type',
-                    'get_in_schedule', 'has_url', 'status')
+                    'get_in_schedule', 'has_url', 'status', 'review_score')
```
drnlm: It may also be useful to have the number of reviews here

wafer/talks/models.py
```diff
@@ -221,6 +222,11 @@ def has_url(self):
 
     has_url.boolean = True
 
+    @property
+    def review_score(self):
+        reviews = [review.total_score for review in self.reviews.all()]
+        return sum(reviews) / len(reviews)
```
drnlm: Shouldn't we guard against the zero division case here?

wafer/talks/models.py
```diff
@@ -221,6 +222,11 @@ def has_url(self):
...
+
```
drnlm: Could we add sorting on review score? I can see that being quite useful in the admin interface to look at cases that need attention.

wafer/talks/models.py
```diff
@@ -268,3 +274,51 @@ class TalkUrl(models.Model):
...
+@python_2_unicode_compatible
+class Review(models.Model):
+    talk = models.ForeignKey(Talk, on_delete=models.CASCADE,
+                             related_name='reviews')
+    reviewer = models.ForeignKey(settings.AUTH_USER_MODEL,
+                                 on_delete=models.CASCADE)
+
+    notes = models.TextField(
+        null=True, blank=True,
+        help_text=_("Comments on the proposal"))
```
drnlm: Would this perhaps be better as a MarkupField, so we can have formatted reviews and all that?

wafer/talks/models.py
```diff
@@ -268,3 +274,51 @@ class TalkUrl(models.Model):
...
+        return self.scores.aggregate(total=models.Sum('value'))['total']
+
+
+@python_2_unicode_compatible
+class ReviewAspect(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __str__(self):
+        return self.name
+
```
drnlm: Perhaps we should add some sort of review weighting here?

I can imagine lots of cases people would want to place more emphasis on a particular aspect during the review process.

wafer/talks/admin.py
```diff
@@ -70,7 +71,18 @@ class TrackAdmin(VersionAdmin, admin.ModelAdmin):
...
 
+class ReviewScoreInline(admin.TabularInline):
+    model = Score
+
+
+class ReviewAdmin(admin.ModelAdmin):
+    list_display = ('talk', 'reviewer', 'total_score')
+    inlines = (ReviewScoreInline,)
+
+
```
drnlm: It would be useful to expose the talk and review update times from reversion somewhere, so we can spot cases where the talk has been updated after a review has been submitted.

wafer/talks/models.py
stefanor: I think that would require putting the review score into a field on the review model. Just including it in `sortable_by` doesn't seem to do anything.

wafer/talks/admin.py
stefanor: Talks aren't currently "reversion"ed. Should we?

wafer/talks/models.py
drnlm: Possibly - I'm more ambivalent about changing notes, but I can see an argument for it.

wafer/talks/admin.py
drnlm: Revisions are created when talks are edited  - see talks/views.py

Reviews should also be handled by reversion.

wafer/talks/models.py
drnlm: On further reflection, what I think we need filters on review score and review count., rather than sorting. As a talk coordinator, I'd want to be able answer questions like "which talks have no reviews yet" or "which talks have at least X reviews and a score above / below Y"  from the admin view and filtering the list will enable that.

wafer/talks/models.py
stefanor: Yeah, it's hard to imagine any kind of pluggable score aggregation that's going to be clean.

Weights could make sense.

wafer/talks/admin.py
gwolf: FWIW as a content team leader on a conference which runs on Wafer, I think something like this could work:

1. If a talk is modified and has not been rated, allow it to be modified. For our purposes, it wouldn't even have to be revisioned.
2. If a talk has been rated but is not yet published, modifying it should be versioned (extra points if versioning means a diff is available). The talk's ratings are marked as "stale", content team members must confirm them or modify them.
3. If a talk has been accepted, it should _not_ be modifiable anymore except by the content team. Authors that want to modify accepted talks are encouraged to mail content@ indicating the changes.

drnlm: 2 (and 1) should happen automatically. Diffs for talks are available from the admin interface. Depending on the user interface we want for reviews, we may need to hook up diff views outside the admin interface, as we've done for pages (perhaps with more than just the "previous version" comparison that the page editing UI supports).

Once the talk status is changed to "Pending", "Accepted" or "Rejected", it is no longer editable by the authors, so that should take care of 3, although we may want more fine-grained control of talk editing permissions than the current "admin access only" model.

wafer/talks/models.py
stefanor:
```
17:48 < gwolf> (coming up with the right set of aspects and the right weighing seems like too byzantine ;-) )
```

stefanor:
I took a shot at UI. The reversion bit doesn't seem to be working at all. Anyone know why?

wafer/talks/views.py
```diff
@@ -63,7 +65,18 @@ def get_object(self, *args, **kwargs):
...
+        user = self.request.user
+
+        context['can_edit'] = talk.can_edit(user)
+
+        can_review = Review.can_review(user)
+        context['can_review'] = can_review
+        if can_review:
+            review = talk.reviews.filter(reviewer=user).first()
+            context['review_form'] = ReviewForm(
+                instance=review, talk=talk, user=user)
```
drnlm: For many smaller conferences, the pool of speakers and the pool of reviewers is going to overlap.
Should we explicitly check that that the reviewer is not an author?

stefanor: Done in 4490164

wafer/talks/templatetags/review.py
```diff
@@ -0,0 +1,23 @@
...
+def reviewed_badge(user, talk):
+    """Returns a badge for the user's reviews of the talk"""
+    if user.is_anonymous():
+        return ''
+    review = talk.reviews.filter(reviewer=user).first()
+    if not review:
+        return ''
+    if review.is_current():
+        return format_html(
+            '<span class="badge badge-success" title="{}">R</span>',
```
drnlm: We already use "R" for not accepted talks. while the styling will be different, it's probably a good idea to not reuse the same letter for different status indications.

wafer/talks/models.py
```diff
@@ -268,3 +286,64 @@ class TalkUrl(models.Model):
...
+    @property
+    def total_score(self):
+        return self.scores.aggregate(total=models.Sum('value'))['total']
+
+    def is_current(self):
+        def last_updated(obj):
+            version = Version.objects.get_for_object(obj).first()
+            return version.revision.date_created
+        return last_updated(self) >= last_updated(self.talk)
```
drnlm: This doesn't work as expected currently, because currently revisions are created for reviews when talks are updated, because reversion's automatic registration adds all related models to the follow attribute.

I think we need to explicitly register Talks and exclude Review from being followed to have this work.

wafer/talks/templatetags/review.py
stefanor: Whoops, missed that.

wafer/talks/models.py
stefanor: Hrm, I didn't notice that in my testing, but the analysis sounds good.

stefanor: Fixed in a37c22b

PR #421, opened 2018-04-14
Hide action buttons when printing pages

stefanor:


hodgestar:
We should also maybe add `d-print-none` to our navigation bar?

PR #426, opened 2018-04-18
Change review values to be 0-centered (-2..2)

stefanor:
This makes them a little easier to deal with.

This had been something I was considering, and I think @gwolf would prefer it.

Additionally, I think using averages everywhere will make adding generic filters for score ranges easier.


hodgestar:
Could you explain the "a little easier to deal with"? I can imagine various reasons, but I don't know the ones that prompted this PR.

stefanor:
> Could you explain the "a little easier to deal with"?

That summation would work almost as well as a mean, for aggregation. But that's probably a very minor point.

PR #429, opened 2018-04-29
Create a Talk Reviewers group with review permissions

stefanor:
Whoops. Missed that one.


drnlm:
Mentoring and Review are, in my mind, different groups. In PyCon ZA, for example, there has traditionally been no overlap between those reviewing talks and those mentoring speakers. 

I think overloading Mentors with the review role is the wrong thing to do. We should add a Review group,  with the view_all_talks permssion and add_review, but it should be a different group from Mentors.
(Probably with access to notes & private notes, although that's not the correct fit for all conferences)

stefanor:
OK, can do that. Of course conferences can customize the permissions, too.

PR #430, opened 2018-04-29
Documentation for Talk Review

stefanor:
There are some unrelated cleanup commits here, but that should be trivially reviewable by commit.

This PR assumes that #426 and #429 are merged. #427 also has docs changes, but those were easy to keep within that PR.


stefanor:
Rebased, as #428 caused a conflict.

PR #435, opened 2018-05-31
Feature/sponsor tag files

drnlm:
Following discussion of #pyconza and #wafer, this implements the idea of managing files via a list of tagged names.

I think the approach is sound, and allows moving a bunch of sponsor handling into the templates.


hodgestar:
Looks good.

Bikeshedding question -- spaces in tag names could be awkward in some circumstances. Should the default tag names be something like "main-logo" and "footer-logo" instead?

If we're happy with the change (which I am) we should document the special tag names. 

@stefanor Do you see anything missing from this for DebConf's use cases?

docs/sponsors.rst
```diff
@@ -27,3 +27,16 @@ Files
...
+------------------------
+
+Uploaded files can be associated with a sponsor and a name in the admin
+interface which can be used with the ``sponsor_tagged_image`` templatetag
+in the templates.
+
+The default wafer sponsor templates expect each sponsor to have an image
+labelled ``main_logo`` for use in the sponsor list.
+
+Wafer also provides an example template block for adding sponsors as a
```
hodgestar: Should we mention the name of the inclusion tag for the blog (``sponsors_footer``) here too?

PR #440, opened 2018-07-30
Hooks for customizing profile pages + styling improvements

stefanor:
DebConf does a fair bit of customization to the profile page, for our registration system.

1. We add links to the right-hand menu
1. We show bursary request status
1. We show accommodation details

Right now, we have to replace the entire profile template. But we can add some hooks for doing the above, by breaking the template up into several included parts. This PR does that.

Also, it breaks each section into a `<section>` so we can put some whitespace around them.


wafer/users/templates/wafer.users/snippets/profile_bio.html
```diff
@@ -0,0 +1,82 @@
...
+                {% blocktrans %}
+                  Registered
+                {% endblocktrans %}
+              </div>
+            {% else %}
+              <div class="alert alert-danger">
+                {% blocktrans %}
+                  <strong>WARNING:</strong>
+                  Talk proposal submitted, but speaker hasn't registered to attend.
+                {% endblocktrans %}
```
drnlm: It seems a bit odd to have this bit of talk related logic here, and not in profile_talks

Perhaps we should move it so everything talk'ish is in one place.

stefanor:
> Changing the sectioning and the CSS classes will mess with styling on PyConZA's site

Yep, this would affect debconf too. We're currently using different CSS classes.

> Currently we mostly use a top-level section with the classes wafer wafer-<kind-of-section> on it and then sometimes a sub-section with classes wafer wafer-<kind-of-section>-<kind-of-subsection>.

Yep consistency is good.

Is the `wafer` class useful for anything? I've always wondered what point it served.

Are you asking for a section around these sections? We do have `{% block container_class %}userprofile{% endblock %}`.

hodgestar:
I'm saying that the top-level section `<section class="wafer wafer-profile">` that used to surround everything is gone and exists in other cases were we have a list of sections.

The wafer class is helpful sometimes for styling all elements inside any wafer sections (e.g. for styling tables in a particular way everywhere, or changing list bullets).

drnlm:
@hodgestar: would re-adding the enclosing `<section class="wafer wafer-profile">` be enough to address your concerns, or do you also want tweaks to the class definitions to the enclosed sections?

hodgestar:
> @hodgestar: would re-adding the enclosing `<section class="wafer wafer-profile">` be enough to address your concerns, or do you also want tweaks to the class definitions to the enclosed sections?

I'm happy with just re-adding the enclosing section.

stefanor:
Rebased, reworked, and this time I left the enclosing section in place.

stefanor:
> this time I left the enclosing section in place

Note that I don't think any other page has nested sections, and with the pyconza2023 styling, that does look a bit weird.

drnlm:
> I'm not convinced that this is ideal for styling.

I think we can just replace the outermost section tag with a div - we mainly want some place to have page-wide css tags for PyConZA

drnlm (approved):

hodgestar (approved):
Looks good.

PR #442, opened 2018-08-01
Group talk by talk type in the talks list

drnlm:
The current talk list doesn't display the talk type, and just sorts the talks by submission order, which can be confusing when several different types are involved.

This groups the list into the different types.

We arguably want to add some sort of "list_title" field to talk types, since it looks a bit odd to use the singular form in the list, and equally odd to use the plural for submissions, and trying to do something automatic will be hard for translation.


hodgestar:
Looks good to me.

We might want to be able to order the talk types at some point (e.g. have keynotes first or something) and perhaps order talks differently within the list (maybe alphabetically?) but that feels like something for another time.

drnlm:
We already have a "order" attribute to control the ordering of talk types, so that can be tweaked.

Having better options for controlling the order of the actual talks would be useful.

Issue #384, opened 2017-10-08
Schedule Items needs to track more metadata


drnlm:
For integration with the video tools, we need to be able to track which items will / are being recorded.

We may actual want to have a few fields here:
video
streamed
etc.

which starts looking more like a list of tags than just an extra field, so we need to think about the exact requirements here a bit.

stefanor:
In DebConf's experience, venues need to have a video flag. Then being videoed becomes the default for a talks in a venue. Possibly the flag should be on slots, not venues. We've had conferences in the past where we moved equipment between rooms during the conference.

Of course, this doesn't guarantee that events will be recorded. You need volunteers to make video happen. So some sort of "request video", "granted video" workflow.

drnlm:
Based on the recent online portions from PyConZA, it would also be useful to be able to add indicators like "don't have the bot announce this"

drnlm:
Although maybe the scheduleitem is the wrong level because they can often be deleted when the schedule is updated - perhaps we need to attach the metadata to the schedulable items?

stefanor:
> Although maybe the scheduleitem is the wrong level because they can often be deleted when the schedule is updated - perhaps we need to attach the metadata to the schedulable items?

I *think* @rhonda would probably appreciate that too

drnlm:
There are a few existing django lbraries for adding arbitary tags which can probably solve most of this - django-taggit seems to be the most actively maintained of those I've looked at, but I haven't done a complete trawl yet,

PR #445, opened 2018-08-07
Add video flag to venues

stefanor:
It's quite possible that not all venues have video coverage. DebConf and PyCon ZA are both examples of this.

DebConf has done hacks like `id <= 3` and `description.contains("has video coverage")` in the past, but actually modelling this data would be a better solution.


hodgestar:
How are other exceptional cases handled -- e.g. video is down in a venue one morning? And how does the video tooling access this? Via the API?

drnlm:
See #384 for some prior discussion about the same issue.

Venue feels like the wrong level of granuality for this, since a venue can easily span multiple days, but not have video for all of them (or all sessions in a single venue on a single day).

stefanor:
> How are other exceptional cases handled -- e.g. video is down in a venue one morning?

We handle this on a talk by talk basis. The content team unsets the video request flag on talks that are not going to be videoed (usually because they're being last-minute scheduled, and the video team can't guarantee the room will be staffed).

> And how does the video tooling access this?

We're actually using it in our templates, e.g. https://debconf18.debconf.org/schedule

https://salsa.debian.org/debconf-team/public/websites/dc18/blob/master/templates/wafer.schedule/schedule_item.html#L12

And the volunteer system hooks into it:
https://salsa.debian.org/debconf-team/public/websites/dc18/blob/master/volunteers/forms.py#L198

> See #384 for some prior discussion about the same issue.

Oh, right, thanks for reminding me of that. Yes, separating video requests from granting video would probably also be useful.

> Venue feels like the wrong level of granuality for this, since a venue can easily span multiple days, but not have video for all of them (or all sessions in a single venue on a single day).

This is true. At DC14, we moved equipment between rooms occasionally (for special events, basically). But this is vary rare because of the set-up time involved.

So, yes, having a video flag on slot is probably more technically correct than on venue. But you'd probably want to seed its value with a value from the venue, because they'd almost always match.

hodgestar:
Could we update the help text to explicitly mention that the flag indicates that items scheduled in this venue will default to being videoed?

I think I'm okay with this landing. I agree that tags provide a much more flexible solution (and I don't think we should abandon that idea), but I'm also wary of creating something too generic for core features (and being videoed seems like a very common feature of conference talks these days).

stefanor:
> We should revisit this after PyConZA, though.

OK.

PR #447, opened 2018-08-14
Feature/use django bakery

drnlm:
Since django-medusa is very dead, this switches to using django-bakery to handle the static site generation.

This produces results almost identical to that from django-medusa on the PyConZA 2017 database. The major differences are the inclusion of cancelled talks and the talk track and type views, which the django-medusa renderer doesn't know about. This also doesn't honour exclude-from-static yet, which it probably needs to learn about.

I also still need to test the handling of ``WAFER_PUBLIC_ATTENDEE_LIST``.


hodgestar:
I'm for merging this and creating issues to remind us of the outstanding niggles -- it's a big step forward over not-having-something-that-will-continue-to-work.

drnlm:
This is now basically feature equivalent to the django-medusa version, and handles WAFER_PUBLIC_ATTENDEE_LIST in the same way on my tests.

Issue #449, opened 2018-10-12
Add a customisable license for videos


drnlm:
We should allow conferences to choose a license for video releases and display it to speakers when they choose whether to allow video for their talks.

Issue #450, opened 2018-10-12
Separate video export from pentabarf and all that


drnlm:
Hacking all the video information into the pentabarf export is not the best idea, despite it's convience.

We should add a specific video info export (json probably makes the most sense) to handle this use case.

PR #451, opened 2018-10-13
Add a WAFER_VIDEO_LICENSE setting

stefanor:
Show the license on the talk submission form, and include the license in the pentabarf XML export.

Fixes: #449

This ignores #450, where we may want to remove the pentabarf bit.


drnlm:
I think we need to also have some way of linking to the full license text, but this certainly better than the current state.

Issue #436, opened 2018-06-21
displays "Page not found" error when user not logged in


dpocock:
If a user is not logged in and they click the link at the bottom of the bursary status email, e.g.

     https://debconf18.debconf.org/users/some_user

the "Page not found" error appears with the text "The page you requested cannot be found"

In this case it should probably show "Unauthorized" and prompt for the user to log in.

hodgestar:
I suspect the end point is designed this way to avoid leaking information about the users that exist to people who are not logged in. Maybe it could be improved though.

dpocock:
could it simply give the "Unauthorized" error for any username, whether the username is valid or not?

stefanor:
The better thing would have been to link to the profile redirect page, @olasd.

hodgestar:
Thinking this through a bit. There are six cases:

User list visibility | User exists & accessible | User exists & not accessible | No such user
-----------------------|---------------------------------|---------------------------------------|------------------
public | 200 | 403 | 400
private | 200 | ? | ?

The private list constraint requires that the two missing values be the same. Returning a 400 means that a person whose session has timed out will see the page not found template. Returning a 403 means that a logged in user will receive a forbidden page when accessing a non-existent user.

Give this, I'm happy to return a 403 for the private user list in both cases since in a sense this is correct -- all information about the user list is forbidden.

Wafer currently doesn't have a base 403 template though, so we should add one.

hodgestar:
Closed by #454.

PR #454, opened 2018-10-16
Bugs/issue 436 return 403 for all profiles

drnlm:
Hopefully addresses #436 

This should return 403 for all user related urls when WAFER_PUBLIC_ATTENDEE_LIST is False and the user isn't logged in.

We are probably too aggressive about returning 404's when WAFER_PUBLIC_ATTENDEE_LIST is True, (see the tests) but I haven't thought through all the implications of changing the behaviour.

### DIFF
--- a/wafer/users/tests/test_views.py
+++ b/wafer/users/tests/test_views.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+# vim:fileencoding=utf-8 ai ts=4 sts=4 et sw=4
+"""Tests for wafer.user.views"""
+
+from django.contrib.auth import get_user_model
+from django.test import Client, TestCase
+from wafer.talks.models import Talk, ACCEPTED
+
+
+class UserProfileTests(TestCase):
+
+    def setUp(self):
+        create = get_user_model().objects.create_user
+        # Create 2 users
+        create("test1", "test@example.com", "test_password")
+        create("test2", "test@example.com", "test_password")
+        user3 = create("test3", "test@example.com", "test_password")
+        talk = Talk.objects.create(title="Test talk", status=ACCEPTED,
+                                   corresponding_author_id=user3.id)
+        talk.authors.add(user3)
+        self.client = Client()
+
+    def test_not_logged_in_private_userlist(self):
+        """We should get 403 for both existing and non-existant users.
+        
+           We can see the user with the accepted talk."""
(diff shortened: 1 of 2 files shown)


drnlm:
Any opinions on this at all?

hodgestar:
I liked @stefanor's suggestion to link to the login / profile redirect page instead (https://github.com/CTPUG/wafer/issues/436#issuecomment-408578538).

Redirecting to a login page with a 403 makes sense for users who are logged out, but returning a 403 for pages that don't exist to people who are logged in seems odd. The previous logic of returning 404s seems better, and I'm happy for people who turn on WAFER_PUBLIC_ATTENDEE_LIST to have to think a bit about which pages they link to in emails.

The new tests are great, so it would be good to keep those if we can.

drnlm:
> I liked @stefanor's suggestion to link to the login / profile redirect page instead ([#436 (comment)](https://github.com/CTPUG/wafer/issues/436#issuecomment-408578538)).

This is far from the only way a user can try access their profile while being logged out, and many of them, such as bookmarks or browser history, are outside the control of the conference.

> Redirecting to a login page with a 403 makes sense for users who are logged out, but returning a 403 for pages that don't exist to people who are logged in seems odd. 
>The previous logic of returning 404s seems better, and I'm happy for people who turn on WAFER_PUBLIC_ATTENDEE_LIST to have to think a bit about which pages they link to in emails.

Did you mean "turn off WAFER_PUBLIC_ATTENDEE_LIST" ? This doesn't change the behaviour for the public list case.

With a private user list, we will deny access to users who are logged in all the time - there is a lot of things non-admin users are not allowed to do (edit other people's profiles, see users without talks, etc).  The issue is that we don't want to allow any sort of enumeration of users based on error code, so we collapse both 404's and 403's into the same error message, regardless of the user's logged in state. Collapsing everything into 404's leads to the complaints we see in #436. I fail to see a compelling reason why returning 404 to a logged in user for an existing page is a better choice  than returning 403 for a non-existent page and contend that collapsing everything to 403 is going to create a better user experience.

There is an argument that we should have different behaviour when the user is an admin, since they can already enumerate the user list via the admin site, and returning 403 instead of 404 is misleading, but that is a future enhancement.

hodgestar:
After reading @drnlm's comment and thinking through this a bit more, I'm happy with this! During my thinking and trying things out, I realized wafer doesn't have a custom 403.html template, so we should add that (in a separate PR probably).

Issue #452, opened 2018-10-13
Talks can't cross midnight


stefanor:
The data model for wafer schedule slots is:
* Day that have a date
* Venue (many:many with Days)
* Slot (many:1 with Venue) has only a time, not a date
* ScheduleItem (many:1 with Slot + Venue)

Because slots are tied to a specific date, a talk can't cross midnight.

The solution here is probably to give Days a start and end datetime, and allow the end to be on a different date. Slots would use datetimes too, and be validated to fit within a Day. It may make sense to rename Day to ScheduleBlock or something.

olasd:
As long as `Slot`s start using `datetime`s instead of having to join them to a `Day` to get one, I'll be happy, I guess. It'll at least reduce the number of silly hax in the volunteer app, as well as in the "schedule in your timezone" stuff that was implemented during DebConf.

stefanor:
Hrm, can this be closed now?

drnlm:
Closing since, with #456, it's possible to have talks cross midnight.

Further enhancements and fixes should be new issues.

PR #456, opened 2018-10-20
Feature/rework schedule days

drnlm:
Addresses #452 

The schedule editor needs some fixes and the "current" and "day" schedule views also need attention, but this demonstrates the approach @stefanor and I discussed after PyCon ZA 2018.


wafer/schedule/models.py
```diff
@@ -12,17 +12,60 @@
...
+       the times for obj2"""
+    if (obj2.end_time >= obj1.end_time and
+        obj2.get_start_time() <= obj1.get_start_time()):
+        return True
+    return False
+
+def overlap(obj1, obj2):
+    """Test if obj1 and obj2 overlap on times"""
+    # obj1 and obj2 are either both ScheduleBlocks or both slots
+    if (obj2.get_start_time() >= obj1.get_start_time() and
```
hodgestar: Alternative implementation ideas:
* We could write `if (obj1.get_start_time() <= obj2.get_start_time() < obj1.end_time): return True` unless there is an odd reason this doesn't work as expected, or
* We could write simply `if (obj1.end_time <= obj2.get_start_time()): return False` if we're happy relying on `obj1.get_start_time() < obj1.end_time`? If we check these two conditions instead, I think we can avoid the call to `include(..)`?

wafer/schedule/models.py
```diff
@@ -12,17 +12,60 @@
...
+        ordering = ['start_time']
 
+    def get_start_time(self):
+        """Helper method to ease validation checks."""
+        return self.start_time
+
+    def __str__(self):
+        if self.start_time.date() != self.end_time.date():
+            return u'%s - %s' % (self.start_time.strftime('%b %d (%a) %H:%M'),
+                                 self.end_time.strftime('%b %d (%a) %H:%M'))
```
hodgestar: @stefanor I'm wondering how DebConf would like this rendered? If one declares that a day ends at 2am the following morning, displaying full datetimes here seems clunky, but I'm not sure what is less clunky and how to handle more complicated cases.

@drnlm If people put multiple blocks all within the same day, the date() == date() check will succeed and they'll all end up having the same day.

Maybe both of these are fine for now and we should just land this and see what to do next though.

wafer/schedule/models.py
```diff
@@ -68,31 +111,41 @@ class Slot(models.Model):
                                                   "slot OR a day and start "
                                                   "time set)"))
 
-    day = models.ForeignKey(Day, null=True, blank=True,
+    block = models.ForeignKey(ScheduleBlock, null=True, blank=True,
```
hodgestar: Now that slots have datetimes, I'm wondering whether it makes sense to drop the link from slots to blocks and simply use the blocks as time range and venue filters for displaying slots with?

wafer/schedule/models.py
drnlm: That's true, but I'm not sure how much effort we should put into preventing that sort of thing. I suspect we'll ultimately need to add some sort "name" field to blocks to override the default in the cases were it doesn't match what the organisers want.

hodgestar:
Just noting for myself that I've reviewed everything up to here and it looks good.

drnlm:
I have now finished tweaking things here,

hodgestar:
@stefanor Does this look like it will meet DebConf's requirements?

hodgestar:
@stefanor Re-pinging! :)

hodgestar:
Looks good to me still post merge of master and related updates.

wafer/schedule/models.py
drnlm: Noting that this was done in e2acebf

wafer/schedule/models.py
drnlm: Noting that this was rewritten and simplified in 157fe90

wafer/schedule/models.py
drnlm: I think adding some sort of custom field to override the default is a decent option, but I'll do that as a separate PR so it's easier to sort out the details

PR #458, opened 2019-03-05
profile page improvements

terceiro:
While I was working on the DebConf19 registration, I ended up making these
changes to 1) link to registration from the profile and 2) hide the "Submit
Talk Proposal" if WAFER_TALKS_OPEN is False.

I do realize this conflicts with #440, and that we will have to override the
entire profile page for DebConf 19 anyway. But this looks like an incremental
change that could be useful enough in wafer upstream.

### DIFF
--- a/wafer/users/templates/wafer.users/profile.html
+++ b/wafer/users/templates/wafer.users/profile.html
@@ -30,12 +30,15 @@
           <ul class="float-right btn-group btn-group-vertical profile-links">
             <li><a href="{% url 'wafer_user_edit' object.username %}" class="btn btn-secondary">{% trans 'Edit User' %}</a></li>
             <li><a href="{% url 'wafer_user_edit_profile' object.username %}" class="btn btn-secondary">{% trans 'Edit Profile' %}</a></li>
-            {% if WAFER_REGISTRATION_OPEN and not profile.is_registered %}
-              {% if WAFER_REGISTRATION_MODE == 'ticket' %}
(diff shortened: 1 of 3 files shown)


wafer/users/templates/wafer.users/profile.html
```diff
@@ -30,15 +30,20 @@
...
+            {% if WAFER_REGISTRATION_OPEN %}
+              {% if WAFER_REGISTRATION_MODE == 'ticket' and not profile.is_registered  %}
                 {% url 'wafer_ticket_claim' as register_url %}
               {% endif %}
+              {% if WAFER_REGISTRATION_MODE == 'custom' %}
+                {% url 'register' as register_url %}
+              {% endif %}
               {% if register_url %}
-                <li><a href="{{ register_url }}" class="btn btn-secondary">{% trans 'Register' %}</a></li>
+                <li><a href="{{ register_url }}" class="btn btn-secondary">{% if profile.is_registered %}{% trans 'Update registration' %}{% else %}{% trans 'Register' %}{% endif %}</a></li>
```
drnlm: "Update registration" isn't a phrase that's accurate for the ticket registration something - maybe something more like "Check registration" covers more cases?

terceiro: in the "ticket" mode, register_url is only set if the user is not already registered, so "Update registration will never be displayed". It will only be displayed in "custom" mode, to users who are already registered.

drnlm: Fair point

PR #464, opened 2019-06-02
Fix Review.__str__()

terceiro:
This reorders the arguments so that the resulting text makes sense
("Review of [TALK] by [REVIWER]")


drnlm:
Thanks

PR #465, opened 2019-06-03
wafer_speaker_contact_details: fix output

terceiro:
This fixes the CSV output for wafer_speaker_contact_details on python3.
I'm not sure sure it this would still be needed under python 2, but I
would assume that a recent enough Django will do the right thing either
way.


stefanor:
Yeah, I bet that predates Python 3.

> but I would assume that a recent enough Django will do the right thing either way.

Don't think this is Django's problem. In Python 2, you are feeding bytes into the csv module, which is writing bytes. In Python 3, it can feed unicode into the file object.

PR #476, opened 2019-08-09
Embed view and Bootstrap tables in the schedule template

andrewshadura:
This PR adds two things.

1. Bootstrap classes for the schedule, making it more usable even without customisations:
Bootstrap tables

2. Embed view, suitable for embedding on another website: it has no headers, footers, navigation elements or edit controls.


wafer/schedule/templates/wafer.schedule/embed_schedule.html
```diff
@@ -0,0 +1,79 @@
...
+  <link href="{% static 'vendor/bootstrap/dist/css/bootstrap.min.css' %}" rel="stylesheet">
+  <link href="{% static 'vendor/@fortawesome/fontawesome-free-webfonts/css/fontawesome.css' %}" rel="stylesheet">
+  <link href="{% static 'vendor/@fortawesome/fontawesome-free-webfonts/css/fa-solid.css' %}" rel="stylesheet">
+  <link href="{% static 'css/wafer.css' %}" rel="stylesheet">
+  {% block extra_head %}{% endblock %}
+</head>
+<body>
+{% block content %}
+<section class="wafer wafer-schedule">
+  <div class="wafer_schedule">
```
stefanor: If we're going to have this view, I suspect the core schedule logic should be factored out into an include or a block.

wafer/schedule/templates/wafer.schedule/full_schedule.html
```diff
@@ -35,7 +35,7 @@ <h1>
         </div>
       {% endif %}
       {% for schedule_day in schedule_days %}
-        <table cellspacing=1 cellpadding=0>
+        <table cellspacing=1 cellpadding=0 class="table table-striped table-bordered">
```
stefanor: This is probably a better default, but I can still imagine wanting to override it, so it's tempting to do this via SCSS rather than classes.

Are `cellspacing=1 cellpadding=0` still needed with these classes?

wafer/schedule/templates/wafer.schedule/embed_schedule.html
andrewshadura: True.

wafer/schedule/templates/wafer.schedule/full_schedule.html
andrewshadura: I have removed that bit for now.

PR #481, opened 2019-08-31
Make some models translatable

andrewshadura:
Make the most visible models translatable.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -51,10 +51,11 @@ def authors_help():
 @python_2_unicode_compatible
 class TalkType(models.Model):
     """A type of talk."""
-    name = models.CharField(max_length=255)
-    description = models.TextField(max_length=1024)
-    order = models.IntegerField(default=1)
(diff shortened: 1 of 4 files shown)


andrewshadura:
ping

wafer/talks/models.py
```diff
@@ -51,10 +51,11 @@ def authors_help():
...
-    name = models.CharField(max_length=255)
-    description = models.TextField(max_length=1024)
-    order = models.IntegerField(default=1)
+    name = models.CharField(_('name'), max_length=255)
+    description = models.TextField(_('description'), max_length=1024)
+    order = models.IntegerField(_('order'),default=1)
     disable_submission = models.BooleanField(
+        _('disable submission'),
         default=False,
         help_text="Don't allow users to submit talks of this type.")
```
drnlm: This should probably be updated as well

drnlm: Can be done in subsequent work, though

PR #485, opened 2019-10-03
Feature/highlight schedule venue

drnlm:
As discussed on IRC, this adds a "highlight-venue" parameter to the schedule, which provides a styling hook to differentiate items in the chosen venue in the schedule view.

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/current.html
+++ b/wafer/schedule/templates/wafer.schedule/current.html
@@ -22,7 +22,11 @@
         <tr>
           <th>{% trans "Time" %}</th>
           {% for venue in schedule_day.venues %}
-            <th><a href="{{ venue.get_absolute_url }}">{{ venue.name }}</a></th>
+            <th
+              {% if venue.pk == highlight_venue_pk %}
+                class="schedule-highlight-venue"
+              {% endif %}
+            ><a href="{{ venue.get_absolute_url }}">{{ venue.name }}</a></th>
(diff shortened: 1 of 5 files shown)


wafer/schedule/templates/wafer.schedule/current.html
```diff
@@ -36,13 +40,18 @@ <h1>{% trans "Happening now" %}</h1>
...
-                {% if item.note == "complete" %}
-                  <td colspan="{{ item.colspan }}" rowspan="{{ item.rowspan }}" class="completed">
-                {% elif item.note == "current" %}
-                  <td colspan="{{ item.colspan }}" rowspan="{{ item.rowspan }}" class="current_active">
-                {% else %}
-                  <td colspan="{{ item.colspan }}" rowspan="{{ item.rowspan }}" class="future">
-                {% endif %}
+                <td colspan="{{ item.colspan }}" rowspan="{{ item.rowspan }}"
+                  {% if item.item.venue.pk == highlight_venue_pk %}
+                    class="schedule-highlight-venue"
```
hodgestar: Could this end up with the class be specified multiple times?

hodgestar: I see from the tests that is does. I would have thought that was not okay in HTML?

drnlm: I misremembered the HTML spec on this - fixed

PR #490, opened 2019-11-07
add per-talk type submission deadline and option for late submissions

terceiro:
Sometime conferences want to continue to receive submissions after the
deadline, and maybe accept them if there is space and they are
interesting enough. This patch displays them as being late submissions
in both the talk list and in the talk page, for the benefit of
reviewers and content admins.

This feature is completely opt-in: if WAFER_TALKS_DEADLINE is not set,
nothing changes.


terceiro:
I have seen that the tests fail on python2 because I'm using unittest.mock. I can easily amend my patch to fallback to the top-level mock module, which is even already in requirements-dev.txt. but there is a larger point which is: should we really still care about python2 at this point?

terceiro:
> should we really still care about python2 at this point?

you know what, nevermind. my point still stands, but I don't want this PR to be about anything other than the actual feature I am proposing.

drnlm:
In general, I like the idea.

However, I am concerned that a single deadline is too simplistic. We have multiple talk types, and may want different deadlines for each of them. For PyCon ZA, for example, we usually close tutorial submissions well before closing talk submissions (which is why we have support for individually disabling sugmissions of each talk type).

terceiro:
yeah, having this be per-track would be great. can you point me to the code that implements that? I skimmed the pyconza2019 repo but couldn't find it.

hodgestar:
One option might be to put the deadline into the TalkType model (if we feel that talk type is the correct way to break up the different deadlines) -- https://github.com/CTPUG/wafer/blob/master/wafer/talks/models.py#L51-L77

terceiro:
argh I confused talk type with track. yes IMO it makes more sense to do this by talk type (and in there indeed there is separate submission flag)

terceiro:
ok, so I reworked my original idea, and now added a submission_deadline field and a "accept late submissions" flag to TalkType. I then replaced all direct usage of disable_submission with a new method in TalkType.objects that fetches all talk types that can have talks submitted to them.

wafer/talks/models.py
```diff
@@ -47,6 +48,21 @@ def authors_help():
     ))
     return ' '.join(text)
 
+class TalkTypeManager(models.Manager):
+
+    def open_for_submission(self, plus=models.Q(pk__lt=0)):
```
hodgestar: I would remove the `plus` parameter and simply allow the caller to do their own `| Q(pk=initial) `.

wafer/talks/tests/test_talk_type.py
```diff
@@ -0,0 +1,54 @@
+def test_open_by_default():
+    from wafer.talks.models import TalkType
```
hodgestar: I assume this strange importing of things inside the test functions was copied from `test_wafer_basic_talks.py`? If so, does anyone remember why it is like that? If not, perhaps we should just try importing things the ordinary way at the top of the module?

wafer/talks/templates/wafer.talks/talk.html
```diff
@@ -108,6 +108,11 @@ <h1>
         {% else %}
           <span class="badge badge-danger">{% trans 'Not accepted' %}</span>
         {% endif %}
+        {% if object.is_late_submission %}
+        {% with submission=object.submission_time %}
+        <span class="badge badge-warning ">{% blocktrans %}Late submission ({{ submission }}){% endblocktrans %}</span>
```
hodgestar: Should we display the submission time somewhere even if the talk was not submitted late?

wafer/talks/tests/test_talk_type.py
drnlm: This pattern is from the first tests we wrote  - I think we needed to work around some odd behaviour with how importing `django.contrib.auth` worked when running the tests at the time, and just never updated it once that was no longer needed.

wafer/talks/models.py
terceiro: done

wafer/talks/tests/test_talk_type.py
terceiro: ack, I moved the imports to the top

wafer/talks/templates/wafer.talks/talk.html
terceiro: good idea

terceiro:
hello, I just addressed all of the comments, and pushed a new version of this PR

terceiro:
nice, thanks! it would be cool to publish a new release, there's been quite some work in master since the last one

PR #493, opened 2019-11-23
Add multisite support

terceiro:
All but the user models are now site-aware, i.e. by default they will
only load objects from the current site, and saving new objects will
automatically assign them to the current site.

This feature is opt-in, i.e. if you have a single conference and do
nothing, everything will just work as before.

A current limitation is that user management is global, and any
permissions given apply to all sites. I believe that addding support for
site-specific permission assignment is possible, and it's in my plans to
tackle that in future work.


drnlm:
Could explain why you think this is solution is better than simply archiving previous conferences and reusing the user data from the last one if user continuity is desired?

hodgestar:
To expand on @drnlm's comment a bit:

We (PyConZA) have found it extremely useful to use wafer's support for generating static copies of a site to archive old conferences. This approach has a number of big advantages -- it's a proper static archive so it doesn't change over time; it means we can upgrade wafer each year without having to worry about an ever growing history of old data to migrate; it means each site is independent so modifying templates and custom code doesn't change or break previous sites.

A comment on the PR itself -- I'm not against better use of Django sites support, but overriding the base model class seems like quite a big price to pay unless there is a very significant gain. Is this a common approach to multisite support?

terceiro:
the use case I want to support is running multiple conferences at the same time.

if you use run one conference at a time, archiving works great fine. actually one thing does not invalidate the other. you can still archive old sites.

terceiro:
> A comment on the PR itself -- I'm not against better use of Django sites support, but overriding the base model class seems like quite a big price to pay unless there is a very significant gain. Is this a common approach to multisite support?

TBH I didn't find anyone else doing it like this. I came up with this to make the multisite support transparent: i.e. as long as the settings are in place, no other code needs to change. of source overriding the base Django model class is a bit scary, but that is why I took care to make that as small as possible

terceiro:
FWIW, the test that is failing on travis has nothing to do with the changes that I did (and the tests all pass for me locally on this branch).

hodgestar:
> the use case I want to support is running multiple conferences at the same time.

Ah, I see. Could you explain a bit how you see the setup looking? E.g. would each site have separate settings.py files? Or just one? And what are the advantages of having these all in the same DB? If seems like it would be better to just run separate wafer instances (it's not a particularly heavy application) and have completely separate databases and everything else (which seems to be the end goal you're pursuing anyway?). What things do you see the two sites sharing?

> TBH I didn't find anyone else doing it like this. I came up with this to make the multisite support transparent: i.e. as long as the settings are in place, no other code needs to change. of course, overriding the base Django model class is a bit scary, but that is why I took care to make that as small as possible

I think there are quite a few difficulties in this approach. E.g. already the `ContentTypes` table and `User` tables are shared and not really fixable. The same would be true for all tables created outside of wafer (e.g. by plugins) and for all other global things that aren't site aware (e.g. settings if you see those being shared).

terceiro:
> > the use case I want to support is running multiple conferences at the same time.
> 
> Ah, I see. Could you explain a bit how you see the setup looking? E.g. would each site have separate settings.py files? Or just one?

just one

> And what are the advantages of having these all in the same DB?

sharing users accounts for example. 

> If seems like it would be better to just run separate wafer instances (it's not a particularly heavy application) and have completely separate databases and everything else (which seems to be the end goal you're pursuing anyway?). What things do you see the two sites sharing?

Ideally my goal is allowing friends who are not python programmers, and/or aren't be as comfortable as I am with sysadmin stuff, be able to instantiate new conferences on their own. for that I want to leave a single wafer instance running, and give them the access to create new sites via the admin interface.

> 
> > TBH I didn't find anyone else doing it like this. I came up with this to make the multisite support transparent: i.e. as long as the settings are in place, no other code needs to change. of course, overriding the base Django model class is a bit scary, but that is why I took care to make that as small as possible
> 
> I think there are quite a few difficulties in this approach. E.g. already the `ContentTypes` table and `User` tables are shared and not really fixable. The same would be true for all tables created outside of wafer (e.g. by plugins) and for all other global things that aren't site aware (e.g. settings if you see those being shared).

yeah well I am aware of the limitations. some of them may be worked around, some won't.

I took special care to make this as unobtrusive as possible to the normal use case of a single conference per instance/db. It shouldn't be noticeable most of the time; the only effect will be that every query will now have "WHERE site_id=1".

hodgestar:
> Ideally my goal is allowing friends who are not python programmers, and/or aren't be as comfortable as I am with sysadmin stuff, be able to instantiate new conferences on their own.

If this is the goal, then there are still quite a few other pieces to solve yes? I suppose at the very least there will need to be separate domain names for every site so that the multisite support would work? I think really conferences would probably also need different settings.py, custom base templates, and separate permission control (these are certainly things that all conferences that have used wafer have needed so far).

Would a better solution not be to make it easier to deploy wafer, so that a simple script or process could create a new instance with minimal intervention? That would keep the sites separate and also be useful to new users wanting to create there first wafer site?

hodgestar:
As a somewhat random example, one could use uWSGI's Emperor mode -- https://uwsgi-docs.readthedocs.io/en/latest/Emperor.html as part of such of a solution.

PR #495, opened 2019-11-23
css: make small improvements to the default stylesheet

terceiro:
These are small things that make quite some difference in the default
user experience.

### DIFF
--- a/wafer/static/css/wafer.css
+++ b/wafer/static/css/wafer.css
@@ -1,3 +1,8 @@
+#main {
+   padding-top: 1em;
+}
+
+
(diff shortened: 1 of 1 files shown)


hodgestar:
@terceiro Woot. These look good to me, but would you mind explaining in a bit more detail what each of the changes is fixing so I know what issue you were seeing and how the CSS change addressed it?

terceiro:
@hodgestar done; split the changes into two commits, each with a reasonable explanation

hodgestar:
Thanks!

PR #496, opened 2019-11-23
.travis.yml: only report coverage over the wafer source itself

terceiro:
i.e. don't count coverage of all of the libraries that we use.


hodgestar:
Woot on the coverage limitation. This PR seems to include some other changes. Could you update it to just include the coverage commits?

terceiro:
done. sorry for the noise, I got confused with quite a few branches

hodgestar:
Np! Thanks for fixing this!

PR #499, opened 2019-11-30
settings: drop USER from PostgreSQL settings

terceiro:
This makes it possible to run against PostgreSQL without having to have
access to the `postgres` user. When USER is omitted, it will try to use
the current username. If a different username is needed, the user can
set the PGUSER environment variable.


terceiro:
nevermind, I realized that this is needed for the travis tests

PR #500, opened 2019-12-02
wafer.settings: don't load localsettings

terceiro:
Based on the documentation (docs/install.rst), the recommended way of
using wafer is in a new Django app that includes wafer settings in its
own. In that case, it's less surprising to let that app handle any local
settings loading, if any is needed. In fact, most apps using wafer that
I came about did their own loading of a localsettings module.


hodgestar:
Looks sensible! Thanks!

PR #501, opened 2019-12-03
wafer.setting: log errors to the console

terceiro:
I'm trying to debug problems with sending email. In this case, the only logging
available being sending email is ... suboptimal. :-)

### DIFF
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -165,9 +165,6 @@ INSTALLED_APPS = (
 
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
 
-# A sample logging configuration. The only tangible logging
-# performed by this configuration is to send an email to
-# the site admins on every HTTP 500 error when DEBUG=False.
 # See http://docs.djangoproject.com/en/dev/topics/logging for
 # more details on how to customize your logging configuration.
 LOGGING = {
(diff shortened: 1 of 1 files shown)


stefanor:
Is that really a default we want? I wouldn't want this in production.

hodgestar:
I think there are three roles this setting can play:
1. Production default
2. Development default
3. Documentation (i.e. an example of how to set up logging)
Currently in practice it is used as 1 by PyConZA. I think this PR helps 2 at the expense of 1. Any ideas on how to fulfill all three better?

drnlm:
Given that we advise people to build an app around wafer, we should consider dropping the default logging section from settings.py entirely, and move the comment to the docs, and add explicit instructions that users should setup their own logging configuration. Maybe also suggesting some common choices?

terceiro:
IMO logging errors to stderr is a reasonable default, even for production. If you are running under apache/mod_wsgi, those get to the apache error log. If you are running e.g. gunicorn in a docker/"cloud native" environment, that's how you are supposed to do logging in the first place.

In general I think it's useful to provide as many sensible defaults as possible, to reduce the amount of boilerplate one has to add to apps using wafer. And, those who have strong opinions about those settings can already easily override them.

stefanor:
Fair enough

drnlm:
There isnt going to be a universally good default, and having too many options enabled by default is also a bad idea. I'm not particulatly convinced that wafer should actually be providing an enabled logging method by default, but having both mail and logging to stderr enabled doesn't seem generally useful.

Also, whatever we have enabled should be discussed in the documentation, and updating / changing the logging setup should be discussed in the installation document.

PR #503, opened 2020-01-10
Expand on the notes field help text

stefanor:
Requested by the DebConf content team

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -6,7 +6,7 @@ from django.urls import reverse
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import lazy
 from django.utils.translation import ugettext, ugettext_lazy as _
-from django.utils.timezone import datetime, now
+from django.utils.timezone import now
 
 import reversion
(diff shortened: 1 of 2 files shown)


wafer/talks/models.py
```diff
@@ -169,7 +169,9 @@ class Meta:
     notes = models.TextField(
         _("notes"),
         null=True, blank=True,
-        help_text=_("Any notes for the conference organisers?  "
+        help_text=_("Any notes for the conference, e.g. why you are in a good "
+                    "position to present on this topic, if it is not "
+                    "clear from your bio already. "
```
hodgestar: I agree that more explanation is needed, but I'm a bit -1 on this wording. It feels too much like "please justify yourself" to me?

stefanor: How about:

> Any notes for the conference? Such as additional background on the topic or presenters that isn't going to be published publicly, special requirements for the event, or thoughts on scheduling.

hodgestar: Sounds good!

PR #507, opened 2020-01-25
Improve wording around the sign up process

terceiro:


wafer/registration/templates/registration/activation_complete.html
```diff
@@ -5,8 +5,16 @@ <h1>{% trans 'Activation complete' %}</h1>
...
+    Your account is now activated!
     Please <a href="{{ login_url }}">log in</a>.
 {% endblocktrans %}
 </p>
+<p>
+{% blocktrans trimmed %}
+    Note that registration for the conference is a separate step.
+{% endblocktrans %}
+{% blocktrans trimmed %}
+    You can now do it after logging in.
```
stefanor: How about: You can now log in and start the conference registration process.

Worth noting: Under quicket conference registration does not necessarily require a wafer account.

terceiro: that's much better, thanks. updated

PR #509, opened 2020-01-31
Talk.video: avoid interpolating string too soon

terceiro:
Interpolating right there forces a lookup at Django startup time, and
then you never get a translation at the talk submissino form.

I could have used format_lazy from django.utils.text, but that would
require changing the actual string and causing churn on all translation
(which are few at the moment, to be fare). This one-off solution makes
it possible to keep the string untouched.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -5,6 +5,7 @@ from django.template.defaultfilters import slugify
 from django.urls import reverse
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import lazy
+from django.utils.text import format_lazy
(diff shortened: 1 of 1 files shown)


drnlm:
Is the complexity this adds really worth avoiding one round of updating translations ?

stefanor:
Aren't these easy enough to fix without involving translators?

terceiro:
It doesn't. I just rewrote it using format_lazy instead. :-)

terceiro:
@stefanor hah I missed your comment while I was rewriting this.

TBH the only translator affected so far is myself, so I'm good with any solution, as long as we agree on one. :-)

PR #512, opened 2020-02-01
setup.py: ship .mo files in releases

terceiro:
Note that this makes it so that building the release files now depends
on gettext.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,5 @@
+from glob import glob
+import subprocess
 import sys
 
 from setuptools import find_packages, setup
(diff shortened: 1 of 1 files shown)


drnlm:
This is something we do need to do, but the current implementation breaks how we install dependencies (see travis), so we'll need to fix that before we can merge this.

terceiro:
ack. I have updated the implementation to allow the process to continue in case of failure to compile the translations.

I'm not super happy about it (e.g. the call to ./manage.py will always happen, on every invocation of setup.py), but it works for now; maybe that's good enough.

there are more ellegant/involved ways of accomplishing this, but they will require more effort.

drnlm:
I'm not a huge fan of this approach, but we can improve on this later.

PR #513, opened 2020-02-01
talk submission form: improve display of talk type/track description

terceiro:
Unless the track description is very short, this makes the <select> box
enormous, often overflowing the screen width. In any cases, good track
names should be self-explanatory so one shouldn't need a description in
the select box items.

### DIFF
--- a/wafer/talks/forms.py
+++ b/wafer/talks/forms.py
@@ -29,6 +29,16 @@ def has_field(model, field_name):
         return False
 
 
+class TalkCategorisationWidget(forms.Select):
+
+    class Media:
+        js = ('wafer/talks/talk-categorization.js',)
+
+    def __init__(self, *args, **kwargs):
+        super(TalkCategorisationWidget, self).__init__(*args, **kwargs)
+        self.attrs["class"] = "talk-categorization"
+
+
 class TalkCategorisationField(forms.ModelChoiceField):
     """The categories that talks can be placed into.
     These are always required, if there are any registered.
@@ -36,6 +46,7 @@ class TalkCategorisationField(forms.ModelChoiceField):
     def __init__(self, model, initial=None, empty_label=None, *args, **kwargs):
         super(TalkCategorisationField, self).__init__(
             initial=initial,
+            widget=TalkCategorisationWidget(),
             queryset=model.objects.all(),
             empty_label=None,
             required=True,
--- a/wafer/talks/static/wafer/talks/talk-categorization.js
+++ b/wafer/talks/static/wafer/talks/talk-categorization.js
@@ -1,0 +1,30 @@
+var $ = (jQuery || django.jQuery)
+
+$(document).ready(function() {
+
+    function formatTalkCategorizationResult(item) {
+        var parts = item.text.split(/:\s*/)
+        var name = parts[0]
+        var description = parts[1]
+        return $([
+            "<div>",
+            "<div>",
+            "<strong>",
+            name,
+            "</strong>",
+            "</div>",
+            description,
+            "</div>"
+        ].join(""))
+    }
+
+    function formatTalkCategorizationSelection(item) {
+        return item.text.split(/:\s*/)[0]
+    }
+
+    $('.talk-categorization').select2({
+        templateResult: formatTalkCategorizationResult,
+        templateSelection: formatTalkCategorizationSelection,
+        dropdownAutoWidth: true
+    })
+})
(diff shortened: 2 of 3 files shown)


drnlm:
I think this is a good idea, but I'm going to hold off on merging to see if we can come up with a good way of addressing Simon's comments.

terceiro:
hello @drnlm and @hodgestar thanks for the feedback. I rewrite this PR to use select2 for those dropdowns. It now hides the description when the dropdown is closed, and displays it with less when the dropdown is open, but with line wrapping and hightlighting the item name.

It looks like this:

track-dropdown

terceiro:
I don't understand why the ci is failing; there are different failures, and the ones I found don't seem related to my changes

wafer/talks/static/wafer/talks/talk-categorization.js
```diff
@@ -0,0 +1,30 @@
...
+    function formatTalkCategorizationSelection(item) {
+        return item.text.split(/:\s*/)[0]
+    }
+
+    $('.talk-categorization').select2({
+        templateResult: formatTalkCategorizationResult,
+        templateSelection: formatTalkCategorizationSelection,
+        dropdownAutoWidth: true
+    })
+})
```
drnlm: We already depend on django-easy-select2 for the schedule - isn't it possible to reuse that for this, rather than adding custom logic ?

wafer/talks/forms.py
```diff
@@ -29,13 +29,24 @@ def has_field(model, field_name):
         return False
 
 
+class TalkCategorisationWidget(forms.Select):
+
+    class Media:
+        js = ('wafer/talks/talk-categorization.js',)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
terceiro: ah! fixed, thanks

wafer/talks/static/wafer/talks/talk-categorization.js
terceiro: it's not.

I read the documentation for both easy-select2 and select2, and the relevant code from select2, and there is no way to pass any non-string data (such as those callback functions) othen than via the select2() call.

there were more than one attempt to add that to select2 (e.g. select2/select2#2957 and select2/select2#4702) but none of them got anywhere.

drnlm:
@hodgestar Does this address your concerns?

wafer/templates/wafer/base_form.html
```diff
@@ -1,12 +1,15 @@
...
 {% load crispy_forms_tags %}
+{% load static from staticfiles %}
 {% block extra_head %}
   {{ form.media.css }}
 {% endblock %}
 {% block content %}
   {% crispy form %}
 {% endblock %}
 {% block extra_foot %}
+  <script type="text/javascript" src="{% static 'admin/js/jquery.init.js' %}">
```
hodgestar: What does this change do?

hodgestar:
@terceiro The new approach looks awesome. I left a question about the manual inclusion of `jquery.init.js` but otherwise looks great.

wafer/templates/wafer/base_form.html
terceiro: it makes any select2 fields work at all. for some reason, jquery was not being loaded in any non-admin pages.

hodgestar: There is this in the base template though -- https://github.com/CTPUG/wafer/blob/master/wafer/templates/wafer/base.html#L37 ?

There is also this weird thing inside the schedule editor -- https://github.com/CTPUG/wafer/blob/master/wafer/static/js/scheduledatetime.js ?

Maybe we need a snippet like that inside talk-categorization.js somehow?

terceiro: the issue is that easy_select2 requires `django` and `django.jQuery` to be defined; see  `easy_select2/static/easy_select2/js/easy_select2.js `.

`django` and `django.jQuery` are defined in `admin/js/jquery.init.js`. It seems like easy_select2 (or at least, its latest version) is only meant for the admin, and we are kind of abusing it in arbitrary forms. or maybe that's just an oversight in easy_select2

terceiro: to clarify: this fixes the __existing__ select2 fields in wafer, not the ones I am adding in this PR (as for those I am calling select2 directly due to the reasons explained above)

hodgestar: Could you add a comment to the template describing why this is here? After that I think I'm happy with the PR.

hodgestar: Also, thanks for sorting this out!

terceiro:
done!

PR #514, opened 2020-02-05
Feature/update travis matrix

drnlm:
Update travis to test (and fail) against recent Django releases, so we have an accurate picture.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,18 @@ python:
    - "3.5"
    - "3.6"
    - "3.7"
+   - "3.8"
 env:
    - DJANGO='django>=1.11.0,<1.12.0' TESTDB=sqlite
    - DJANGO='django>=1.11.0,<1.12.0' TESTDB=postgres
    - DJANGO='django>=2.0,<2.1.0' TESTDB=sqlite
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,14 +28,52 @@ matrix:
...
+      - python: "3.8"
+        env: DJANGO='django>=1.11.0,<1.12.0' TESTDB=postgres
+      - python: "3.8"
+        env: DJANGO='django>=2.0,<2.1.0' TESTDB=sqlite
+      - python: "3.8"
+        env: DJANGO='django>=2.0,<2.1.0' TESTDB=postgres
+      - python: "3.8"
+        env: DJANGO='django>=2.1,<2.2.0' TESTDB=sqlite
+      - python: "3.8"
+        env: DJANGO='django>=2.1,<2.2.0' TESTDB=postgres
```
stefanor: I can't help wondering if these work, even if Django doesn't explicitly support it. I don't see any commits that claim to be related to 3.8 support, in their git HEAD.

drnlm: While it's not officially supported, several projects run their tests on Django 2.1 and Python 3.8, so it's currently doable to at least some extent.

PR #521, opened 2020-03-11
settings: re-introduce localsettings under wafer/

terceiro:
Even though everyone Should™ be using wafer from other apps, it's useful
to run plain wafer, from the wafer sources, to debug issues in isolation.

### DIFF
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -325,3 +325,8 @@ BAKERY_VIEWS = (
     'wafer.users.views.UsersView',
     'wafer.users.views.ProfileView',
 )
+
+try:
+    from wafer.localsettings import *
+except LoadError:
+    pass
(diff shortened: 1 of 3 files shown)


wafer/settings.py
```diff
@@ -325,3 +325,8 @@
     'wafer.users.views.UsersView',
     'wafer.users.views.ProfileView',
 )
+
+try:
+    from wafer.localsettings import *
```
stefanor: How about doing this the other way around. So that it only impacts dev, not library consumers?

`manage.py --shell localsettings`

drnlm: Would that work? AFAICS manage.py will set the `DJANGO_SETTINGS_MODULE` environment variable to  the settings.py file regardless.

Maybe we need documentation section explaining how to setup a local settings file and best use it?

terceiro: nah, I guess anyone hacking on wafer needs to be aware of how Django works anyway. maybe we should just drop this

drnlm:
Closing per last comment - if a better idea comes along, we can open a different PR

PR #524, opened 2020-04-13
Feature/minor cleanups

drnlm:
CLeaning up some dead python2 compatibility code and also fixing a number of pep8 warnings while we're here.

### DIFF
--- a/wafer/schedule/tests/test_admin.py
+++ b/wafer/schedule/tests/test_admin.py
@@ -209,7 +209,7 @@ class SlotListFilterTest(TestCase):
         if time:
             return SlotStartTimeFilter(None, {'start': time}, Slot, self.admin)
         else:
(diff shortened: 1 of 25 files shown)


wafer/schedule/tests/test_admin.py
```diff
@@ -209,7 +209,7 @@ def _make_time_filter(self, time):
         if time:
             return SlotStartTimeFilter(None, {'start': time}, Slot, self.admin)
         else:
-            return SlotStarTimeFilter(None, {'start': None}, Slot, self.admin)
+            return SlotStartTimeFilter(None, {'start': None}, Slot, self.admin)
```
hodgestar, line 212: Woops. :?

drnlm: Indeed.

That we haven't tripped over this previously suggests that we need to look at the tests here more carefully at some point.

PR #526, opened 2020-04-13
Feature/replace easy select2

drnlm:
We simplify our select2 handling a bit by replacing django-easy-select2 with the inbuilt autocomplete in the admin interface and using django-select2 for the non-admin forms.

This allows us to remove the awkward hack of needing jquery-admin outside the admin interface to make django-easy-select2 work.


wafer/users/admin.py
```diff
@@ -1,11 +1,36 @@
...
 
+class AuthorAutocompleteView(AutocompleteJsonView):
+    """We override the default so we can use render_author
+       in the admin display"""
+
+    def get(self, request, *args, **kwargs):
+        """Override request to change the response object"""
+        response = super().get(request, args, kwargs)
+        # This isn't very efficient, but we don't want to duplicate
+        # the specialised logic in AutocompleteJJsonView either
```
hodgestar, line 22:
```suggestion
-        # the specialised logic in AutocompleteJJsonView either
+        # the specialised logic in AutocompleteJsonView either
```

PR #529, opened 2020-06-01
Replace Debian SSO with GitLab SSO

stefanor:
Debian's client-certificate based SSO mechanism is being retired,
replaced by SSO through Salsa, a GitLab instance.

https://lists.debian.org/debian-project/2020/04/msg00000.html

This is very similar, but not 100% identical to the GitHub implementation.

This branch is built on #528. The relevant change is in the the last commit.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -26,10 +26,6 @@ Wafer's settings
     The name of the Django cache backend that wafer can use.
     Defaults to ``'wafer_cache'``.
 
-``WAFER_DEBIAN_NM_API_KEY``
-    The API Key for ``nm.debian.org``'s API, used by Debian SSO.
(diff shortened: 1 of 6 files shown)


drnlm:
Given this change, doesn't it make sense to switch over to one of the more standard Django libraries for Oauth stuff (Google searches suggest that Social Auth (https://github.com/python-social-auth) is the current consensus choice) rather than continuing to roll our own?

stefanor:
I was wondering that too. We didn't use that for GitHub, at the time, because we'd still have to have put enough glue on top of it, that it wasn't a buying us a whole lot. These OAuth handshakes are fairly simple.

I think the biggest gain would be maintenance (i.e. avoiding #528).

PR #536, opened 2020-06-25
Represent times in localtime for schedule administration

OdyX:
Schedule slot's administration is very confusing with a TIME_ZONE != 'UTC'. The times are stored in a TZ-aware manner, so it's just a display consideration; let's address this in the model's `__str__` representations.

### DIFF
--- a/wafer/schedule/models.py
+++ b/wafer/schedule/models.py
@@ -3,6 +3,7 @@ from django.db import models
 from django.db.models.signals import post_save, post_delete
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
+from django.utils.timezone import localtime
 
 from wafer.pages.models import Page
 from wafer.snippets.markdown_field import MarkdownTextField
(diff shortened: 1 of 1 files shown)


hodgestar:
Thank you for the PR!

I'm not sure I fully understand the current situation and so I'm not quite sure that `localtime(...)` is the correct approach here. I think currently wafer honors the TIME_ZONE setting in settings.py, e.g. https://github.com/CTPUG/pyconza2020/blob/master/settings.py#L160, and displays the schedule times using this timezone. If this is true, then I think the current behaviour is good.

If localtime tries to display the time in the browser / client's time, I think that wouldn't be that useful for schedule editing, because presumably the conference organizers will have a canonical shared timezone that they use when thinking about the times of the schedule items.

Do you disagree? Or am I wrong about any of these?

OdyX:
What I have observed is that on `/admin/schedule/slot`, on our Django+wafer project, configured with `TIME_ZONE = 'Europe/Zurich'` , without this patch, the `__str__` values as displayed in the columns 1 and 3 are displayed in UTC (currently, a `- 2 hours` offset), while the editable field in column 4 is displayed in local time. While editing the schedule slots on `/admin/schedule/slot/{id}/change`, the descriptions in "previous slot" are also in UTC, while the time fields are (rightfully) in localtime.

So, my understanding is that the times are rightfully input as localtime (as any editor would expect), but stored as "time + TZ". Upon re-editing, Django does its magic and will display (and accept) times in localtime, and store them as timezone-aware timestamps. When "manually" displaying these times through `strftime()`, Python (not Django) will "omit" the timezone part, and display the UTC times, which is confusing.

My patch just wraps the time conversions towards localtime; inspired from https://stackoverflow.com/questions/11909071/using-strftime-on-a-django-datetime-produces-a-utc-time-in-the-string . The alternative is to rely on Django filters (see the SO link), but clearly, Python's strftime will omit the TimeZone, which is _very_ confusing.

hodgestar:
Thanks for the explanation! It would be better to have a function that always displays the default time zone, perhaps, but Django doesn't seem to have a built in function for that, so localtime looks good.

Thanks for the patch -- merging!

PR #537, opened 2020-07-02
Add a setting to make talks visible to authenticated users

OdyX:
In the context of our internal conference, authentication is only possible for employees from our domain, and we want them to be able to see all the proposed talks. This patch is needed to achieve this at minimum cost.

### DIFF
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -261,6 +261,9 @@ WAFER_HIDE_LOGIN = False
 # Set this to False to disable talk submissions
 WAFER_TALKS_OPEN = True
 
+# Talk submissions are visible to all attendees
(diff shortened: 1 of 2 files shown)


drnlm:
Since we already have the custom "view_all_talks" permission, shadowing it with a config setting seems like a sub-optimal way of solving this problem.

We do need more options for user customisation to address cases like this, though - maybe specifying a list of default groups users are added to?

OdyX:
I couldn't easily find where I could make all registered users a member of a group *upon registration*. If you have a pointer, I'd be happy to update my patch.

drnlm:
We've added support for this with #540 - let us know if that isn't enough for your use case.

PR #538, opened 2020-07-11
talk admin: fix review count

terceiro:
The number of reviews needs to be normalized by the number of review
aspects.

### DIFF
--- a/wafer/talks/admin.py
+++ b/wafer/talks/admin.py
@@ -91,7 +91,7 @@ class TalkAdmin(CompareVersionAdmin):
     def get_queryset(self, request):
         qs = super(TalkAdmin, self).get_queryset(request)
         qs = qs.annotate(
-            review_count_annotation=models.Count('reviews'),
+            review_count_annotation=models.Count('reviews', distinct=True),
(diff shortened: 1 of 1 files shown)


wafer/talks/admin.py
```diff
@@ -90,8 +90,9 @@ class TalkAdmin(CompareVersionAdmin):
 
     def get_queryset(self, request):
         qs = super(TalkAdmin, self).get_queryset(request)
+        review_aspect_count = ReviewAspect.objects.count()
         qs = qs.annotate(
-            review_count_annotation=models.Count('reviews'),
+            review_count_annotation=models.Count('reviews') / review_aspect_count,
```
hodgestar, line 95: This doesn't seem right -- there should be only one review for each review, although there may be multiple scores for each review. Maybe something else has gone wrong or perhaps you could explain a bit more why you think there is a review for each aspect and not only a score for each aspect?

hodgestar:
Ah, I see that is an issue with how combining multiple aggregates works in Django -- https://docs.djangoproject.com/en/3.0/topics/db/aggregation/#combining-multiple-aggregations.

New fix looks good to me!

PR #539, opened 2020-07-11
talks admin: add actions for changing talk status in bulk

terceiro:


terceiro:
I pushed this on top of #538 because otherwise there would probably be a merge conflict

wafer/talks/admin.py
```diff
@@ -88,10 +128,20 @@ class TalkAdmin(CompareVersionAdmin):
...
+        mark_not_accepted,
+        mark_provisional,
+        mark_accepted,
+    ]
+
     def get_queryset(self, request):
         qs = super(TalkAdmin, self).get_queryset(request)
         qs = qs.annotate(
-            review_count_annotation=models.Count('reviews'),
+            review_count_annotation=models.Count('reviews', distinct=True),
```
stefanor, line 144: Looks like this slipped in from #538

wafer/locale/pt_BR/LC_MESSAGES/django.po
```diff
@@ -32,7 +32,7 @@ msgstr "Nos últimos 7 dias"
 msgid "In the last 30 days"
 msgstr "Nos últimos 30 dias"
 
-#: wafer/compare/admin.py:141
+#: wafer/compare/admin.py:140
```
stefanor, line 35: This is why I tend to strip out context from po files

wafer/talks/admin.py
```diff
@@ -71,6 +80,37 @@ def has_delete_permission(self, request, obj=None):
         return False
 
 
+def mark_talks(value, fname):
+    def f(modeladmin, request, queryset):
+        for talk in queryset:
+            talk.status = value
+            talk.save()
```
stefanor, line 87: I think this can be `queryset.update(status=value)`

terceiro: I tried that at first, but with that I get a very strange error:

django.core.exceptions.FieldError

django.core.exceptions.FieldError: Cannot resolve keyword 'review_score_annotation' into field. Choices are: _abstract_rendered, abstract, authors, corresponding_author, corresponding_author_id, kv, notes, private_notes, reviews, scheduleitem, status, submission_time, talk_id, talk_type, talk_type_id, title, track, track_id, urls, video, video_reviewer

this is probably because the queryset contains annotations

stefanor: Right

PR #540, opened 2020-07-12
WIP: Feature/allow default groups

drnlm:
This is an alternative take on solving the problem raised in #537 , allowing a set of default groups to which new users will be added.

I like the general idea, but I'm not convinced that specifying the default groups in the settings file is the right approach - it feels like something that should be managed as part of the admin.

Thoughts and/or opinions?


hodgestar:
I guess one could create a model called `DefaultGroups` that references the groups. I'm not sure it's a huge improvement, although it does perhaps make the option to change the default groups more visible and saves some sysadmin fiddling.

I'm happy with the current approach -- we can always make the solution fancier if needed.

PR #541, opened 2020-08-01
Several improvements/fixes to the schedule editor

terceiro:


wafer/static/css/wafer.css
```diff
@@ -70,6 +70,19 @@ ul.profile-links li {
     background-color: white;
 }
 
+#wafer-edit-schedule-bucket .nav-tabs a.active {
+  font-weight: bold;
+}
```
stefanor, line 75: These tabs still look pretty awful. It'd be really nice to put them in a visual tab / block of some sort.

image

However, this is certainly an improvement, so won't block this PR on that.

wafer/schedule/templates/wafer.schedule/edit_schedule.html
```diff
@@ -119,6 +119,9 @@ <h2>{% trans 'Bucket' %}</h2>
                     id="talk{{ talk.talk_id }}"
                     data-toggle="tooltip" data-placement="left"
                     title="{{ talk.title }}"
+                    {% if talk not in talks_unassigned %}
+                    hidden
```
stefanor, line 123: I think that should be indented, to match the rest of the template style

terceiro: yeah, fixed

wafer/static/css/wafer.css
terceiro: fixed with some spacing.

bucket-nav

I tried to make it look like actual tabs, but I couldn't figure out how yet. I think a good chunk of that html needs to be rewritten.

stefanor: Probably :( And this sections below factored out, because there's a lot of duplication.

PR #542, opened 2020-08-05
penta_schedule: don't mark non-talks as released

terceiro:


drnlm:
Why this change?

AFAICS from the original pentabarf schedule code, released should only be false if the event has not been added to the public schedule. Since we generate our version from the public schedule, I don't see why we should ever have an entry with released set to  False.

wafer/schedule/templates/wafer.schedule/penta_schedule.xml
```diff
@@ -105,8 +105,7 @@
                       {% else %}
                         <description/>
                       {% endif %}
-                      <released>True</released>
-                      <license>{{ WAFER_VIDEO_LICENSE }}</license>
```
hodgestar, line 109: Why is the license tag being removed?

stefanor:
> released should only be false if the event has not been added to the public schedule

Released is a video release. I don't think it has anything to do with being on the public schedule.

wafer/schedule/templates/wafer.schedule/penta_schedule.xml
stefanor: It does go hand in hand with being released

drnlm:
> Released is a video release. I don't think it has anything to do with being on the public schedule.

Maybe - it's been a long time since I looked at the pentabarf stuff. But as conferences can and do release video for non-talk events, and have non-talk events for which there won't be video, this change still doesn't seem helpful.

stefanor:
> can and do release video for non-talk events

Yeah, the question here is whether that should default to yay/nay. Before having to be overridden in the system that's consuming this data. The argument here is that no video is more likely to be accurate.

> and have talk events for which there won't be video

That's handled by:
1. Declaring a venue as non-video, or
2. Not setting the released flag on an individual talk.

drnlm:
> > can and do release video for non-talk events
> 
> Yeah, the question here is whether that should default to yay/nay. Before having to be overridden in the system that's consuming this data. The argument here is that no video is more likely to be accurate.

The current default is sometimes wrong, switching to a different default that is sometimes wrong is not an improvement. If this is a problem that requires fixing, it should be fixed, not tweaked to something differently incorrect.

stefanor:
In the DebConf case, the special events that need video are always an event (talk) I think. But I guess that's not the case for pyconza.

hodgestar:
Could we solve this by looking at the URLs attached to the talk or page? Or should we perhaps think about having some shared attributes between talks and pages for handling the cases like this where we want to ask the same question about both types of object?

terceiro:
I will try to propose something not hardcoded in a new PR

PR #544, opened 2020-08-15
Filter nested TalkUrls

stefanor:
They weren't being filtered in the nested router, `/talks/api/talks/7/urls` would return all URLs, not just the URLs for talk 7.

Seems to have been caused by an inheritance order bug.

### DIFF
--- a/wafer/talks/tests/test_views.py
+++ b/wafer/talks/tests/test_views.py
@@ -717,6 +717,16 @@ class TalkUrlsViewSetTests(TestCase):
             self.mk_result(url_a), self.mk_result(url_b),
         ])
 
+    def test_list_talk_urls_nested_filtering(self):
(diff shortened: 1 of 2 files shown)


wafer/talks/urls.py
```diff
@@ -10,7 +10,7 @@
 
 talks_router = router.register(r'talks', TalksViewSet)
 talks_router.register(
-    r'urls', TalkUrlsViewSet, basename='talks-urls',
+    r'urls', TalkUrlsViewSet, basename='talks-url',
```
hodgestar, line 13: Should we make this `talk-urls` like the class name? It seems a bit odd to use url singular for a view that returns a list of URLs?

stefanor: So, I don't actually know what this is used for. Quite happy to drop that commit.

PR #545, opened 2020-08-15
Add a public field to TalkUrl

stefanor:
It's sometimes useful to have non-public URLs attached to a talk. E.g. the Video Conference URL that the speaker will use to present the talk.

This is often useful, but sometimes not *all* of the URLs, so let's make it easy to build templates that can display just the public URLs.


hodgestar:
The intention of the docstring was to say that the fields would not be editable by the talk authors. They're there for the conference organizers to fill in. The intention wasn't to say they wouldn't be viewable (probably that could have been clearer).

hodgestar:
What is the goal of the new `public` flag? If someone doesn't want the URL to be visible to the public, can it just be left out instead? And why would one want to hide a URL? Presumably a URL is already publicly accessible?

stefanor:
> The intention of the docstring was to say that the fields would not be editable by the talk authors. They're there for the conference organizers to fill in. The intention wasn't to say they wouldn't be viewable (probably that could have been clearer).

Aha. Happy to restore it and tweak it for clarity, there.

> What is the goal of the new public flag?

Our use-case is URLs for Jitsi rooms, that we want to create for each talk and share with some people but not everyone. See   https://salsa.debian.org/debconf-team/public/websites/wafer-debconf/-/merge_requests/93
So, Metadata that we attach to talks that we don't show the whole world. Could use other things for that, too, like KV, but this did seem like the right mechanism. At least from my interpretation of the comment...

wafer/talks/models.py
```diff
@@ -346,15 +346,13 @@ def can_review(self, user):
 
 @reversion.register()
 class TalkUrl(models.Model):
-    """An url to stuff relevant to the talk - videos, slides, etc.
-
-       Note that these are explicitly not intended to be exposed to the
-       user, but exist for use by the conference organisers."""
+    """An url to stuff relevant to the talk - videos, slides, etc."""
```
hodgestar, line 349: Could we update this to:
```
""" A means for attaching relevant URLs to the talk (e.g. videos, slides).

    These URL associations are explicitly not intended to be edited by the author of the talk, but
    rather by the conference organizers.

    URLs with public set to False are not publicly visible.
"""
```

wafer/talks/models.py
```diff
@@ -346,15 +346,13 @@ def can_review(self, user):
...
 
     description = models.CharField(max_length=256)
     url = models.URLField()
     talk = models.ForeignKey(Talk, related_name='urls',
                              on_delete=models.CASCADE)
+    public = models.BooleanField(default=False)
```
hodgestar, line 355: Given that the previous behaviour was equivalent to `public=True` should the default not be True?

stefanor: Sounds good to me

hodgestar (approved):

PR #547, opened 2020-09-07
Add an API for tickets and ticket types

hodgestar:


wafer/tickets/views.py
```diff
@@ -132,3 +140,19 @@ def import_ticket(ticket_barcode, ticket_type, email):
...
     else:
         log.info('Ticket registered: %s. Unclaimed', ticket)
+
+
+class TicketTypesViewSet(NestedViewSetMixin, viewsets.ModelViewSet):
+    '''API endpoint that allows ticket types to be viewed or edited.'''
+
+    queryset = TicketType.objects.all().order_by("id")
+    serializer_class = TicketTypeSerializer
+    permission_classes = (DjangoModelPermissions, DjangoModelPermissionsOrAnonReadOnly)
```
drnlm, line 150: a) Do we really both permission here? DjangoModelPermissionsOrAnonReadOnly is supposed to include everythng DjangoModelPermissions allows.

b) Do we gain anything by exposing the ticket types to anonymous access?

drnlm (approved):

wafer/tickets/views.py
hodgestar: Re a) You're right, we don't.
Re b) I was thinking that ticket types are public information so there was no reason to hide access to them, but thinking more now the full list is usually not public, so I've dropped the anonymous read access.

PR #548, opened 2020-09-08
template: allow overriding wafer static assets

terceiro:
This allows conference apps to to provide the assets in some way other
than by copying the wafer package.json and using npm, e.g. 1) by
pointing to them in a CDN, or 2) by reusing assets provided by OS
packages, or 3) embedding copies of the assets.

### DIFF
--- a/wafer/templates/wafer/base.html
+++ b/wafer/templates/wafer/base.html
@@ -7,10 +7,12 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  {% block wafer_css %}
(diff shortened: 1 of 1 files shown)


hodgestar:
I think having these blocks is potentially a good idea, because it gives people options, but for the use cases you listed wouldn't a custom Django static storage backend be the canonical solution? E.g. https://github.com/etianen/django-s3-storage.

terceiro:
sure.

I'm not sure how to proceed; do you want to amend the commit message?

hodgestar:
@terceiro Mostly I wanted to see if you still wanted this change after my comment, but I think I will just merge it since it seems useful. :)

terceiro:
> @terceiro Mostly I wanted to see if you still wanted this change after
> my comment, but I think I will just merge it since it seems useful.

thanks!

PR #550, opened 2020-09-16
Bugfix/previous slot validation

drnlm:
Currently, trying to change the end_time of Slot A when it's the previous slot of Slot B will always fail, because the overlap check doesn't handle the case correctly.


hodgestar (approved):
Looks good, but is there some reason the new logic doesn't belong in the `overlap` function?

drnlm:
> Looks good, but is there some reason the new logic doesn't belong in the `overlap` function?

overlap is also used to check for overlaps in the schedule blocks, which don't have the same concept of 'previous' involved.

hodgestar:
Woot.

PR #552, opened 2020-09-26
Make schedule item validators configurable

terceiro:


drnlm:
Given how this changes behaviour, this should also be mentioned in the "Adding additional schedule validation" section in the schedule docs.

PR #554, opened 2020-10-10
Feature/rework Travis test matrix

drnlm:
Python 3.5 is EOL, so we now longer need to test on it, and can drop the promise of supporting it.


hodgestar:
Should we add 3.9 while we're here?

drnlm:
> Should we add 3.9 while we're here?

We probably should add that, and add Django 3.1.

hodgestar (approved):

PR #555, opened 2020-10-10
Restrict Django to >2.0 <3.0, add dependencies from 2020 ZA Pycon site

kimvanwyk:


setup.py
```diff
@@ -5,7 +5,11 @@
 from setuptools import find_packages, setup
 
 REQUIRES = [
-    'Django>=1.11',
+    'Django>=2.0, <3.0',
+    'psycopg2-binary',
```
drnlm, line 9: I don't think that we want psycopg-binary as a hard dependency - sqlite works fine for development, and we can theoritically run on other databases. It's sensible for pyconza2020, where we are explicitly running the site against postgres, but seems overly specific for wafer

drnlm (changes requested):

setup.py
kimvanwyk: Makes good sense. Do we want to the 3 mdx packages? I would imagine they're generally useful to more than just pyconza2020.

drnlm: I'm undecided on those. There are a lot of possible markdown extensions (Debcof adds mdx-linkify, for example, which we don't use), so it's hard to argue why some should be automatically installed while others don't.  Perhaps a recommended list in the install instructions rather than as hard requirements would be better option?

kimvanwyk: That seems reasonable - little point in forcing installation of packages that may not be used. I'll take them out of this PR, so this PR is just about the changes to make a fresh install work.

drnlm (approved):

PR #562, opened 2020-10-21
Use named arguments in translatable strings

drnlm:
manage.py makemessages justifiably complains about strings with multiple unnamed arguments as being difficult to translate, so we should not do that.


andrewshadura:
Probably also worth `sed`ding the existing translations to automagically replace those.

Issue #377, opened 2017-10-03
Limit which speakers are listed on the speaker page


drnlm:
When using talk types for several different events, one may not want the submitters for all the types to show up on the speakers page.

For instance, PyConZA uses talk types for Open Space topics and Demos, and the submitters of these should probably not be grouped with the other speakers.

Organising the speakers page by type. instead of purely alphabetically, might also be an option worth looking at.

PR #567, opened 2020-11-09
Feature/group speakers by talk type

drnlm:
Following a discussion in during the PyCon ZA 2020 sprints, this groups the speakers on the speaker list by talk type, rather than purely alphabetically - this makes it possible to highlight / emphasis keynote speakers or other such adjustments

This does mean speakers with multiple accepted submissions are listed in each section - this is probably fine.

It also simplifies fixing #377, as adding a "hide talk type" can now just hide the suitable list, without more complex logic.

### DIFF
--- a/wafer/talks/views.py
+++ b/wafer/talks/views.py
@@ -243,7 +243,8 @@ class Speakers(BuildableListView):
             user__talks__status='A').distinct().prefetch_related(
             'user').order_by('user__talks__talk_type',
                              'user__first_name',
-                             'user__last_name').annotate(
+                             'user__last_name',
+                             'user__username').annotate(
             talk_type=F('user__talks__talk_type__name'))
         bytype = groupby(speakers, lambda x: x.talk_type)
         context['speaker_rows'] = {}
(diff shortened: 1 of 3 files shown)


hodgestar (approved):

wafer/talks/views.py
```diff
@@ -239,8 +241,14 @@ def get_context_data(self, **kwargs):
         context = super(Speakers, self).get_context_data(**kwargs)
         speakers = UserProfile.objects.filter(
             user__talks__status='A').distinct().prefetch_related(
-                'user').order_by('user__first_name', 'user__last_name')
-        context["speaker_rows"] = self._by_row(speakers, 4)
+            'user').order_by('user__talks__talk_type__name',
```
hodgestar, line 244: I imagine we'll want to be able to select the ordering of the talk types by something other than name, but happy to leave that for a future PR.

drnlm: I've changed this to just sort on talk type (which we can control via the order attribute), so it will match the ordering on the talks list.

hodgestar: Sounds good.

hodgestar:
Approved again. :)

PR #568, opened 2020-11-10
Add django generated pot file

drnlm:
Add the pot file to simplify life for translators


hodgestar (approved):

hodgestar:
Not that we need more CI tasks, but perhaps it would be a good idea to add a CI check that the .pot file is up to date? Not necessary for this PR, just a thought.

drnlm:
We do need to work out some processes / checks around i18n at some point.

PR #572, opened 2020-11-12
Feature/switch to GitHub actions

drnlm:
This replaces travis with github actions.


drnlm:
I'm not happy with the repeated code, but as github doesn't currently support yaml anchors, I'm not sure there's a great way to deal with this.

The duplicated code comes from two sources
The postgres / sqlite tests  - we could combine these and use the matrix option to decide the test, but that doesn't allow us to skip setting up the postgres container for the sqlite test runs, and as that step adds 30s to each test run, skipping it when possible seems worthwhile

The allowed failure sections - there isn't a great way to handle this with the current matrix options - adding them as include options is very verbose and not maintainable. I also like the ability to add 'allowed-failure' to the name which this approach allows.

drnlm:
If we merge this, there's also a disucssion to be had of whether we should use github's branch protection options to require passing checks? It might be good just to help deal with the move from travis' reporting to the action's much more verbose list of checks.

hodgestar (approved):
Approved!

The name for a job can be set to something like "name: ${{ matrix.python-version }}-${{ matrix.django-version }}" if we want nicer names.

There might be a way to avoid some duplication using "if: ${{ ... }}" but it's not documented whether this can be used to turn services on or off and I suspect it can't at the moment.

Happy for this to be merged though and for improvements to happen in future PRs if the duplication gets annoying.

hodgestar:
> If we merge this, there's also a disucssion to be had of whether we should use github's branch protection options to require passing checks? It might be good just to help deal with the move from travis' reporting to the action's much more verbose list of checks.

Is your worry that it's hard to see whether a branch is "green" or not because of all the ticks and crosses reported by GitHub actions? I agree that I can't see whether the CI failures on this branch are allowed or not.

hodgestar:
> Is your worry that it's hard to see whether a branch is "green" or not because of all the ticks and crosses reported by GitHub actions? I agree that I can't see whether the CI failures on this branch are allowed or not.

Another option might be to split the "allowed failures" into a separate workflow file so that they're more clearly distinguished in the UI.

hodgestar:
Example workflow from someone else tackling similar problems: https://github.com/MSP-Greg/puma/blob/allow-failure/.github/workflows/non_mri.yml

drnlm:
> > Is your worry that it's hard to see whether a branch is "green" or not because of all the ticks and crosses reported by GitHub actions? I agree that I can't see whether the CI failures on this branch are allowed or not.
> 
> Another option might be to split the "allowed failures" into a separate workflow file so that they're more clearly distinguished in the UI.

I tried this on with moving django-markitup to use github actions - it only really influences the display on the Actions pane, and doesn't make it particularly clear which job is associated with which workflow in PR UI.

drnlm:
> Example workflow from someone else tackling similar problems: https://github.com/MSP-Greg/puma/blob/allow-failure/.github/workflows/non_mri.yml

Another approach is to use the step-level 'continue-on-error' - see for example https://github.com/invenia/PDMatsExtras.jl/pull/5

That results in the PR checks and actions being listed as success, but the failures are marked as annotations in the Actions Pane.

See https://github.com/drnlm/wafer/actions/runs/370042042 and https://github.com/drnlm/wafer/pull/1 for how this ends up looking.

That's makes failures a bit more hidden than I'd like, but it's probably better than the current result.

drnlm:
I've simplified the json file by combining the allowed failure and not allowed cases.

I've also switched to using the step-level 'continue-on-error', as I think that's the best of the different options for handling the 'test against stuff we don't expect to pass yet' cases, although it's not ideal.

.github/workflows/django.yml
```diff
@@ -0,0 +1,91 @@
...
+        ports:
+          # Maps tcp port 5432 on service container to the host
+          - 5432:5432
+
+    name: Postgres - Python ${{ matrix.python-version }}, Django ${{ matrix.django-version }} (Allowed Failures - ${{ matrix.django-version == '3.1.0' || matrix.django-version == 'https://github.com/django/django/archive/master.tar.gz' }} )
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+        django-version: ['2.2.0', '3.0.0', '3.1.0', 'https://github.com/django/django/archive/master.tar.gz']
```
stefanor, line 30: Any reason not to use the string `master` here? we're having to repeat this URL in several places, instead we could just have it once in the install step.

drnlm: Because I didn't revisit this after deciding to split the installation into two conditional steps, rather than always using the value from the matrix.

PR #589, opened 2021-06-02
Handle PermissionDenied when building Users

stefanor:
We hide all user pages behind 403, when WAFER_PUBLIC_ATTENDEE_LIST is
disabled, to avoid information leakage
(d3ccc4cab2c7b0bd63510b0c32ecbc66d3beb527)


drnlm:
Could you update the comment at the top of `build_object` as well, since that's no longer accurate?

stefanor:
Was wondering whether I needed to catch both exceptions or just PermissionDenied. I don't think we'd ever get 404s here?

drnlm:
That's probably correct, given the other changes.

drnlm (approved):

PR #590, opened 2021-06-02
Disallow problematic usernames

stefanor:
Ideally we'd probably customize the auth model, but we still haven't done that.

A user broke a debconf site static rendering, by having a username of `.` and no public talks. That resulted in all other user profiles created up to that point, being deleted. Good job, user :)

Let's protect ourselves from this sort of thing, in the future.

### DIFF
--- a/wafer/registration/validators.py
+++ b/wafer/registration/validators.py
@@ -6,4 +6,4 @@ def validate_username(username):
     if username.startswith('.'):
         raise ValidationError('usernames cannot start with a "."')
     if username in ('index.html', 'page'):
-        raise ValidationError('Not an acceptable username')
+        raise ValidationError('This username is not available')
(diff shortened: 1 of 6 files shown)


wafer/registration/validators.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+from django.core.exceptions import ValidationError
+
+
+def validate_username(username):
+    """. and .. lead to problematic URLs and static file names"""
+    if re.match(r'^\.+$', username):
+        raise ValidationError(
+            "usernames cannot consist of only .")
```
drnlm, line 10: Shouldn't this also mention .. in the error, or say something like 'only .'s' ?

drnlm: Also, shouldn't we block leading .'s entirely? While not as problematic as the directory entries, the creation of arbitrary  hidden filenames can lead to some odd behaviours

stefanor: That sounds reasonable, yes.

drnlm (approved):

PR #600, opened 2021-06-13
Markdown: safe_mode is gone, bleach output instead

stefanor:
Run the HTML from markdown through bleach, to protect against XSS.

### DIFF
--- a/wafer/markdown.py
+++ b/wafer/markdown.py
@@ -1,0 +1,10 @@
+import bleach
+from bleach_allowlist import markdown_tags, markdown_attrs
+from markdown import markdown
+
+
+def bleached_markdown(text, **kwargs):
+    """Try to avoid XSS by bleaching markdown output"""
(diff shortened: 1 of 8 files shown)


wafer/markdown.py
```diff
@@ -0,0 +1,10 @@
+import bleach
+from bleach_allowlist import markdown_tags, markdown_attrs
+from markdown import markdown
+
+
+def bleached_markdown(text, **kwargs):
```
stefanor, line 6: Maybe this should live in django-markitup?

drnlm: I'm not sure we want to build this into django-markitup directly. It's aiming to be renderer agnostic and so this would add a number of dependencies that not all consumers will want.

It certainly worth adding as an example to the django-markitup README, though.

drnlm (approved):

PR #601, opened 2021-06-13
Make it easier to inherit from TalkView

stefanor:
If one wants a per-talk page for something, it's convenient to be able to subclass TalkView.

However, the canonicalization function then bits you, and you have to jump through hoops to avoid it.

Make it easy to customize the canonicalization.

### DIFF
--- a/wafer/talks/views.py
+++ b/wafer/talks/views.py
@@ -80,10 +80,15 @@ class TalkView(BuildableDetailView):
             raise PermissionDenied
         return object_
 
+    def canonical_url(self):
+        '''Return the canonical URL for this view'''
+        return self.object.get_absolute_url()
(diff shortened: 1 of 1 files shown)


hodgestar (approved):

hodgestar:
@stefanor Could you explain a bit more how the custom per-talk page setup you're using works? Just wondering if there are further adjustments we could make to Wafer.

stefanor:
@hodgestar: https://salsa.debian.org/debconf-team/public/websites/wafer-debconf/-/merge_requests/118 (which I don't think I'm going to merge, I suspect it will be thrown away in favour of a separate entirely client-side JS app). Look at `TalkQuestionView`.

PR #602, opened 2021-06-13
A markdown extension that allows referencing staticfiles

stefanor:
It can be useful to reference static assets managed by `django.contrib.staticfiles` in Markdown text.  e.g. to link to a large video that we want a high cache TTL on.

DebConf has been carrying this since 2019 (although I don't think we've used it since), but it really seems more generically useful than our needs.

Not sure if wafer really wants to grow a contrib, or not, though.

### DIFF
--- a/wafer/contrib/mdx_static.py
+++ b/wafer/contrib/mdx_static.py
@@ -1,0 +1,28 @@
+import re
+
+from django.contrib.staticfiles.templatetags.staticfiles import static
+
+from markdown.extensions import Extension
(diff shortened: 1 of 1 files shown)


drnlm:
While this looks useful, my opinion is this is better spun off as it's own thing that can be added if required (like we have for mdx_variables), rather than including it directly in wafer.

stefanor:
Right, https://github.com/CTPUG/mdx_staticfiles

PR #605, opened 2021-06-20
Improve talk listing

terceiro:
It looks like this on DebConf21:

talks-table

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -38,8 +38,17 @@ jobs:
       continue-on-error: ${{ matrix.django-version == '3.1.0' || matrix.django-version == 'main' }}
       run: |
         python -m pip install --upgrade pip
-        pip install psycopg2
         pip install -r requirements.txt -r requirements-dev.txt
+    - name: 'Install psycopg2 for Django < 3.1'
+      # Django < 3.1 is incompatible with psycopg2 2.9
+      # See https://github.com/psycopg/psycopg2/issues/1293
+      if: "matrix.django-version == '2.2.0' || matrix.django-version == '3.0.0'"
+      run: |
+        pip install 'psycopg2<2.9'
+    - name: 'Install psycopg2'
+      if: "matrix.django-version != '2.2.0' && matrix.django-version != '3.0.0'"
+      run: |
+        pip install psycopg2
     - name: Install Django Release
       run: |
         pip install django~=${{ matrix.django-version }}
--- a/wafer/talks/templates/wafer.talks/talks.html
+++ b/wafer/talks/templates/wafer.talks/talks.html
@@ -5,44 +5,65 @@
 {% block content %}
 <section class="wafer wafer-talks">
   {% regroup talk_list by talk_type.name as grouped_talks %}
-  {% for type, talks in grouped_talks %}
-    {% if type %}
-      <h1>{{ type }}</h1>
-    {% else %}
-      <h1>{% trans 'Talks' %}</h1>
-    {% endif %}
-    <div class="wafer list">
-      {% for talk in talks %}
-        <div>
-          {% if not talk.cancelled and not talk.accepted and talk.is_late_submission %}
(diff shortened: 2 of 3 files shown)


terceiro:
FWIW the ci failure seems to due to https://github.com/psycopg/psycopg2/issues/1293.

wafer/talks/templates/wafer.talks/talks.html
```diff
@@ -5,44 +5,63 @@
...
-      {% empty %}
-        <p>{% trans 'No talks accepted yet.' %}</p>
+  <div class="table-responsive">
+    <table class='table table-striped'>
+      {% for type, talks in grouped_talks %}
+        <thead class='thead-dark'>
+          <tr>
+            <th>
+              {% if type %}
+                {{type}}
```
stefanor, line 15: Spaces inside the curly braces, please.

wafer/talks/templates/wafer.talks/talks.html
```diff
@@ -5,44 +5,63 @@
...
+                {% endif %}
+                {% reviewed_badge user talk %}
+                {% if talk.withdrawn %}
+                  <del><a href="{{ talk.get_absolute_url }}">{{ talk.title }}</a></del>
+                {% else %}
+                  <a href="{{ talk.get_absolute_url }}">{{ talk.title }}</a>
+                {% endif %}
+              </td>
+              {% if languages %}
+                <td>{{talk.get_language_display}}</td>
```
stefanor, line 50: And here.

stefanor:
Looks nice, I like it.

wafer/talks/templates/wafer.talks/talks.html
terceiro: gah I always forget about this :-p fixed

wafer/talks/templates/wafer.talks/talks.html
terceiro: fixed.

wafer/talks/templates/wafer.talks/talks.html
```diff
@@ -5,44 +5,63 @@
...
+                <td>{{ talk.get_language_display }}</td>
+              {% endif %}
+              <td>
+                {% with authors=talk.get_authors_display_name %}
+                  <span class="author">{{ authors }}</span>
+                {% endwith %}
+              </td>
+            </tr>
+          {% endfor %}
+        {% else %}
```
stefanor, line 59: Django has a `{% for x in y %} ... {% empty %} ... {% endfor %}` syntax that you can use here.

terceiro: done.

.github/workflows/django.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       continue-on-error: ${{ matrix.django-version == '3.1.0' || matrix.django-version == 'main' }}
       run: |
         python -m pip install --upgrade pip
-        pip install psycopg2
+        pip install 'psycopg2<2.9'
```
hodgestar, line 41: Could we add a comment here linking to the bug and explaining that we can drop this requirement once we drop support for Django 2.2?

I guess the nicest solution is to only use `pyscopg2 < 2.9` for the Django 2.2 tests, but let's not do that in this PR.

hodgestar (approved):

.github/workflows/django.yml
terceiro: 3.0 is also affected. I updated the PR to restrict the pin to Django < 3.1

hodgestar: Hmm. If this also affects 3.0, we should probably tell the Django and psycopg2 developers in the bug you linked to -- psycopg/psycopg2#1293 -- since 3.0 is a lot further from end of life.

drnlm: Django 3.0 wasn't an LTS release, so it hit EOL in April

We need to update the set of tested Django versions, although that can probably wait until we add python 3.10 to the mix as well.

hodgestar (approved):

PR #606, opened 2021-07-05
Talk.review_score: do not exclude zeros

terceiro:
This make the review score calculation in Talk.review_score at least
provide the same value as the one in the admin (which is calculated via
an SQL aggregate).

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -309,7 +309,7 @@ class Talk(models.Model):
     @property
     def review_score(self):
         # Overridden in admin, to allow sorting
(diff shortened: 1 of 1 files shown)


stefanor:
> do not exclude zeros

do not *include* zeros?

stefanor (approved):
Yep, that logic looks closer to a SQL AVG()

hodgestar (approved):

hodgestar:
@terceiro Thanks!

hodgestar:
I guess we should also add a test for this, but since there wasn't a test before, we can do that in a separately.

PR #607, opened 2021-07-05
Emit frab compatible XML

stefanor:
While `pentabarf.xml` is a largely undocumented format, frab has a defined schema: https://c3voc.de/wiki/schedule

This is a fairly minimal change to bring our pentabarf XML format into line with frab's schema.
We could carry both side-by-side, but they're close enough that it probably makes sense to stick with a well defined schema.
Tools will have to be changed to support this, but there aren't that many of them, it should be manageable.

The reason for this was to add support for https://eventfahrplan.eu/ as a client app.

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
+++ b/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
@@ -120,13 +120,7 @@
                         <license>{{ WAFER_VIDEO_LICENSE }}</license>
                       </recording>
                     {% endif %}
-                    <conf_url>{{ items.item.get_url }}</conf_url>
-                    {# It's useful to have the full url available in the xml file. The pentabarf.xml format isn't that well #}
-                    {# standardised, so we add our own full_conf_url tag to accomodate this requirement #}
-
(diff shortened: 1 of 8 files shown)


wafer/schedule/templates/wafer.schedule/penta_schedule.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 {% load i18n %}
 <schedule>
+  <generator name="wafer" version="{{ wafer_version }}" />
+  <version>{# FIXME: We have no schedule versions #}</version>
```
drnlm, line 5: It's not clear to me from the frab documentation what version is supposed to be. Is this something we should update whenever the schedule is edited? Or something else entirely?

stefanor: > Is this something we should update whenever the schedule is edited

That's my guess.

I'm thinking of implementing something like that (probably in django cache), to allow us to have ETAGs and handle `If-Modified-Since` requests against the schedule.

drnlm (approved):

PR #609, opened 2021-07-18
schedule editor: display talk notes as title

terceiro:
Displaying the talk title again in the title is pointless since the
title is already the main element displayed. Showing the notes makes it
easier for the people doing the scheduling to see e.g. notes about
scheduling restrictions etc.


hodgestar:
@terceiro Thanks. If I understand what you're seeing, the title is displayed both in the draggable item and as the tooltip for that item? I agree that the notes are a much more sensible thing to display in the tooltip.

I'm a bit worried that the `div.title` attribute is used somewhere else though. E.g. the schedule JS has this line `e.target.innerHTML = data.getAttribute('title');` which seems to suggest that the title of the draggable is also used to fill in some part of the schedule during editing?

terceiro:
yes, I think I overlooked this. I will send a better version

PR #610, opened 2021-07-22
Double <abstract> tag

stefanor:
Whoops, I trusted our XSD check to validate the schedule, but we didn't have any talks scheduled, only pages.

So, add a talk, and fix the bug.

### DIFF
--- a/wafer/schedule/tests/test_views.py
+++ b/wafer/schedule/tests/test_views.py
@@ -1467,6 +1467,7 @@ class NonHTMLViewTests(TestCase):
         venue2 = Venue.objects.create(order=2, name='Venue 2')
         venue1.blocks.add(day1)
         venue2.blocks.add(day1)
+        venue1.blocks.add(day2)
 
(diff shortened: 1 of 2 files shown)


drnlm (approved):

wafer/schedule/tests/test_views.py
```diff
@@ -1491,6 +1492,13 @@ def setUp(self):
         for index, item in enumerate(items):
             item.slots.add(slots[index // 2])
 
+        user = get_user_model().objects.create_user('john', 'best@wafer.test',
+                                                    'johnpassword')
+        talk = Talk.objects.create(title="Test talk", status=ACCEPTED,
+                                   corresponding_author_id=user.id)
+        talk_item = ScheduleItem.objects.create(venue=venue1, talk_id=talk.pk)
+        talk_item.slots.add(slots[4])
+
```
drnlm, line 1501: We have this logic scattered across a few different schedule tests now. We really should unify that at some point - probably by using the helpers we have for the talk tests.

stefanor: I'll take a shot at it in a follow-up.

PR #611, opened 2021-07-23
Tests: Use common test and user creation fixture functions

stefanor:
Use `create_user()` and `create_talk()` where possible in tests, eliminating some boilerplate.

Also, mock out a few libravatar calls that can slow down tests and require Internet connectivity.

### DIFF
--- a/wafer/talks/tests/fixtures.py
+++ b/wafer/talks/tests/fixtures.py
@@ -7,8 +7,9 @@ def create_talk_type(name):
     return TalkType.objects.create(name=name)
 
 
-def create_talk(title, status, username, talk_type=None):
-    user = create_user(username)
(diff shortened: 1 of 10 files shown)


wafer/talks/tests/fixtures.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def create_talk_type(name):
+    """Create a talk type"""
+    return TalkType.objects.create(name=name)
+
+
+def create_talk(title, status, username=None, user=None, talk_type=None):
+    if username:
+        user = create_user(username)
```
drnlm, line 12: Should we check if both user and username are None and give a more informative error message in that case?

(probably should check for both set as well)

drnlm (approved):

PR #613, opened 2021-07-24
Support client-side caching of schedules

stefanor:
Store a schedule version in the cache, and handle conditional HTTP requests for schedule views.


hodgestar (approved):
It might be slightly odd to have the schedule version change whenever the cache goes away, but anything else would be more complicated.

+1 on merging.

PR #619, opened 2021-08-29
Count gaps in ScheduleItem duration

stefanor:
DebConf has gaps in the Slots between talks, to keep a reasonable-looking schedule while still having clearly marked gaps between talks.

However, the talk length is displayed incorrectly for talks that span across a gap. This had the knock-on effect of disallowing pre-recorded video uploads for talks that were longer than (talk_length - gaps) while shorter than talk_length.

Correct the calculation and add tests.

### DIFF
--- a/wafer/schedule/models.py
+++ b/wafer/schedule/models.py
@@ -333,26 +333,24 @@ class ScheduleItem(models.Model):
     def get_duration(self):
         """Return the total duration of the item.
 
-           This is the sum of all the slot durations."""
-        # This is intended for the pentabarf xml file
-        # It will do the wrong thing if the slots aren't
-        # contigious, which we should address sometime.
-        slots = list(self.slots.all())
-        result = {'hours': 0, 'minutes': 0}
-        if slots:
-            for slot in slots:
-                dur = slot.get_duration()
-                result['hours'] += dur['hours']
-                result['minutes'] += dur['minutes']
-            # Normalise again
-            hours, result['minutes'] = divmod(result['minutes'], 60)
-            result['hours'] += hours
-        return result
+           This is the time from start to end, including any discontinuities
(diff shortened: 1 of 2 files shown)


hodgestar:
I would argue that the problem here is there isn't actually a gap in these talks. However, insert discussion about the pain of having many tiny ScheduleItems for gaps that are largely unused.

I'm not super sure what the right solution is, but I don't making `duration = end - start` the default is it. Perhaps there could be a flag somewhere that says that there isn't a gap? Or a way to insert a the gap into the list of slots properly?

Open to other suggestions -- the "how to indicate short breaks between talks" has been an ongoing Wafer niggle, largely because at the start we kept things simple by declaring the gap as part of the talk from the point of view of the schedule.

One additional constraint is that the scheduling framework is already pretty complex, so it would be nice not to make it too much more complex if we can. :/

stefanor:
> I'm not super sure what the right solution is, but I don't making duration = end - start the default is it

Can you expand on that. From my PoV, this is a simplification, rather than adding complexity.

Unless there's a use-case for having discontinuous schedule-items, where you want to count the discontinuities. I couldn't think of one earlier, but I'd imagine something like an all-day workshop that spans lunch, but you want to carve out a hole for lunch, but use a single Talk for both morning and afternoon.

However, wafer currently disallows scheduling items like that, with the `find_non_contiguous` validator. (DebConf has to disable this to allow discontinuous events that span breaks, like this.

> Perhaps there could be a flag somewhere that says that there isn't a gap?

That could make sense, I can see 3 potential configurations:
1. No discontinuities permitted (Status quo)
2. Discontinuities permitted, but excluded from talk length (workshop example above)
3. Discontinuities permitted, and included in talk length (DebConf)

PR #622, opened 2021-09-24
Specify folder to install the django.mo file in.

hodgestar:
This prevents the warning:
```
warning: install_data: setup script did not provide a directory for 'wafer/locale/it/LC_MESSAGES/django.mo' -- installing right in 'build/bdist.linux-x86_64/wheel/wafer-0.12.0.data/data'
```
when running `python setup.py bdist_wheel`.

See https://docs.python.org/3/distutils/setupscript.html#installing-additional-files.

However, I'm not sure what the right folder to specify is. Does anything use `django.mo` directly?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,9 @@ setup(
     include_package_data=True,
     install_requires=REQUIRES,
     dependency_links=SOURCES,
-    data_files=compile_translations(),
+    data_files=[
+        ('locale', compile_translations()),
+    ],
     setup_requires=[
         # Add setuptools-git, so we get correct behaviour for
         # include_package_data


stefanor:
> Does anything use django.mo directly?

Yeah, gettext does.

I think it needs to be installed into a locale directory within a Django app, to be found.

stefanor (approved):
That *looks* correct, but I haven't tested the result.

drnlm (approved):
I did a local smoke test with the result of this change, and translations are found as expected, so this looks to be doing the right thing.

hodgestar:
@drnlm Thanks!

PR #623, opened 2021-09-28
Use spaceless to reduce strange whitespace in the sponsor template

drnlm:
We're fairly lax about using Django's options for stripping whitespace from our templates, which can cause issues with styling stuff. This makes the sponsor block less full of random newlines and various spacing.

### DIFF
--- a/wafer/sponsors/templates/wafer.sponsors/sponsors_footer.html
+++ b/wafer/sponsors/templates/wafer.sponsors/sponsors_footer.html
@@ -1,10 +1,10 @@
 {% load sponsors %}
 <div id="wafer-sponsors" class="container">
-  {% spaceless %}
-    {% for package in packages %}
-      {% if package.sponsors.exists %}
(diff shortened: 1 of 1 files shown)


wafer/sponsors/templates/wafer.sponsors/sponsors_footer.html
```diff
@@ -1,21 +1,23 @@
...
-                 title="{{ sponsor.name }}">
-          {% else %}
-            <span class="nologo">{{ sponsor.name }}</span>
-          {% endif %}
-        </a>
-      {% endfor %}
-      <br>
-    {% endif %}
-  {% endfor %}
+  {% spaceless %}
```
hodgestar, line 3: Possibly the spaceless block only needs to be around the `<a ...>...</a>` tag? I think have some whitespace around the rest makes some sort of semantic sense (i.e. its white space between sponsors and around things like the br tag).

hodgestar (approved):

PR #624, opened 2021-09-30
Switch to slots.set to support recent Django behaviour

drnlm:
This fixes the current breakage of the schedule editor "Direct assignment to the forward side of a many-to-many set is prohibited."

Works correctly with local testing with Django 3.0, but we should add better tests to this API at some point.


hodgestar (approved):
Thanks!

Issue #608, opened 2021-07-13
Submitting a talk review via the talk page may update the talk status without creating a revision.


hodgestar:
In [TalkReview.form_valid](https://github.com/CTPUG/wafer/blob/6b0cafefff466cb33e92ac4732531f23c9bc26be/wafer/talks/views.py#L234-L237) the talk status is updated if a first review is added and the talk is not yet under review, but no talk revision is created for the update.

This doesn't break anything, but creates a confusing gap in the talk revision audit trail.

drnlm:
Fixed by #626

PR #625, opened 2021-10-09
Feature/add review form tests

drnlm:
As a step towards fixing #608 , this adds some testing around reviews and also fixes a minor bug with the form validation settings

PR #626, opened 2021-10-09
Bugfix/add revision to talk review status change

drnlm:
Following on from #625, this fixes issue #608.


wafer/talks/views.py
```diff
@@ -231,13 +231,20 @@ def form_valid(self, form):
...
             talk.status = UNDER_CONSIDERATION
             with revisions.create_revision():
                 revisions.set_user(self.request.user)
                 revisions.set_comment("Status changed by review process")
                 talk.save()
+            # We also change the date to apotentially changing the talk to
+            # ensure is_current behaves as expected
+            talk_version = Version.objects.get_for_object(talk).order_by('-pk').first()
+            version.revision.date_created = talk_version.revision.date_created
+            version.revision.save()
```
stefanor, line 247: Urgh, this is a little hacky. Can't we update the status before we save the review to avoid having to hack the talk?

stefanor: Yeah, that change works. More code, but less race-y.

PR #630, opened 2022-05-27
Explicitly convert schedule times to event local time

stefanor:
When using a DB that doesn't store TZ information (like sqlite) we
were rendering the schedule in the system's local timezone, not the
conference's timezone.

This confuses some tools that ignore timezones (like sreview).

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
+++ b/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 {% load i18n %}
 {% load tz %}
-{% localtime on %}
 <schedule>
   <generator name="wafer" version="{{ wafer_version }}" />
   <version>{{ schedule_version }}</version>
@@ -129,4 +128,3 @@
     </day>
   {% endfor %}
 </schedule>
-{% endlocaltime %}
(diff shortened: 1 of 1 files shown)


wafer/schedule/templates/wafer.schedule/penta_schedule.xml
```diff
@@ -20,7 +22,7 @@
     <acronym>{{ WAFER_CONFERENCE_ACRONYM }}</acronym>
   </conference>
   {% for page in schedule_pages %}
-    <day date="{{ page.block.start_time.date.isoformat }}" start="{{ page.block.start_time.isoformat }}" end="{{ page.block.end_time.isoformat }}" index="{{ forloop.counter }}">
+    <day date="{{ page.block.start_time|localtime|date:"Y-m-d" }}" start="{{ page.block.start_time|localtime|date:"c" }}" end="{{ page.block.end_time|localtime|date:"c" }}" index="{{ forloop.counter }}">
```
drnlm, line 25: Does localtime here do anything, given that 'localtime on' is already set?

stefanor: The `localtime on` I think has the effect of localizing date and time objects that attempt to render. But if it's already transformed to a string, it's too late.

drnlm: As I read the django documentation, the only difference between the locatime filter, and the 'localtime on' setting is the support for naive datetime objects, which we shouldn't have if 'load tz' is included. Do you have an example where leaving it out gives incorrect results?

stefanor: Yeah, not needed. That was just belt and braces. Removed!

drnlm (approved):

PR #632, opened 2022-06-04
Feature/update test matrix

drnlm:
We've fallen behind in testing stuff a bit - this updates the test matrix so we're at least covering (and failing) on more recent stuff.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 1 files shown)


stefanor:
I wish the allowed failures still showed as failures :(

drnlm:
They are listed as such in the annotatons section on the actions result page - see https://github.com/CTPUG/wafer/actions/runs/2439394886 for example

I welcome suggestions on how to do this better within the limitations of github actions.

stefanor (approved):
Yeah, I don't know of any better way

PR #635, opened 2022-06-15
Feature/drop django nose

drnlm:
Drop django-nose, since it's unmaintained and breaks on python 3.10

This also enables warnings when running the tests, so deprecations are easier to spot.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -74,7 +74,7 @@ jobs:
       env:
         TESTDB: postgres
       run: |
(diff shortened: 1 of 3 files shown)


hodgestar:
Ah, apparently the correct incantation is `coverage run --source='.' manage.py test myapp` so maybe we can add it back easily and be happy? I'm happy for this to merged as is though if you prefer.

drnlm:
Added

The coverage numbers are fairly different from nose's, though, which is a topic for further investigation.

PR #636, opened 2022-06-15
Feature/fix django4 warnings

drnlm:
This fixes the various warnings about stuff removed in django 4  - this is enough to allow the test suite to run if our problematic dependencies are minimally patched.

### DIFF
--- a/wafer/schedule/tests/test_admin.py
+++ b/wafer/schedule/tests/test_admin.py
@@ -237,7 +237,6 @@ class SlotListFilterTest(TestCase):
         TestFilter = self._make_time_filter('11:00')
         # Check lookup details
         lookups = list(TestFilter.lookups(None, self.admin))
-        print(lookups)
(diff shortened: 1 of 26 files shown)


wafer/schedule/tests/test_admin.py
```diff
@@ -237,7 +237,6 @@ def test_time_filter_lookups(self):
         TestFilter = self._make_time_filter('11:00')
         # Check lookup details
         lookups = list(TestFilter.lookups(None, self.admin))
-        print(lookups)
```
hodgestar, line 240: Doh. :)

hodgestar (approved):
Renaming `url` to `re_path` seems like a really odd decision on Django's part. Yes, I get that it's actually constructing a pattern that matches a part of a path, but `re_path` is hardly better. If they felt so strongly about renaming it they could at least have gone to the effort of picking a name that was better than the previous one, like maybe `route`? Oh well. :)

PR #637, opened 2022-06-16
Bugfix/fix test discovery

drnlm:
Django's native test runner uses unittest's discovery approach, which meant we weren't finding tests that nose found.

This tweaks things so we're running all the tests again.


hodgestar (approved):
Woot!

PR #638, opened 2022-06-27
Replace crispy forms automatic form_tag with explict tags

drnlm:
It can be useful to extend the talk form with additional form content, either from additional apps or for client-side validated fields like a "I've have read the FAQ" checkbox. Crispy form's automatic tag makes that harder than it needs to be, and disabling it is the documented method for handling this use case.


hodgestar (approved):
Looks good!

PR #648, opened 2022-09-30
nav: allow redefining background style

terceiro:


hodgestar:
Is the existing class `navbar` not sufficient for styling?

drnlm:
Given that bootstrap is moving away from direct theme specification (navbar-light has been deprecated, navbar-dark has been reworked), we should probably update to current bootstrap and remove some of the hard-coded theme choices rather than mess around too much with adding extra customisation that isn't going to be useful long term.

stefanor:
I wonder if this can just be handled in some SCSS?

terceiro:
yeah it can probably be handled in some other way, but it would laso be nice if the template didn't hardcode navbar-dark bg-dark (which will force me to undo/revert all the styling done by those classes)

stefanor:
Yes, defaulting to dark is probably not useful for customizing

PR #663, opened 2023-06-07
Improve the 'no talks submitted' display for talks and speakers

drnlm:
The current 'No talks accepted' pages are missing bits or blank - this adds a bit more theming so they hopefully look less ugly.

### DIFF
--- a/wafer/talks/templates/wafer.talks/speakers.html
+++ b/wafer/talks/templates/wafer.talks/speakers.html
@@ -33,6 +33,17 @@
         </div>
       {% endfor %}
     </div>
+  {% empty %}
+    <!-- We show an empty entry wo a) there's some content here and
+                                   b) it's easier to check some of the styling -->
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Looks good to me.

I see some tests failed. Not sure if it's related yet.

hodgestar:
I see there is lots of `django.template.exceptions.TemplateDoesNotExist: bootstrap4/field.html`.

drnlm:
> I see there is lots of `django.template.exceptions.TemplateDoesNotExist: bootstrap4/field.html`.

We either need to restrict crispy-forms to an old version, add cripsy_boostrap4 as a depedncy (although there are questions about how long this will be supported given that bootstrap has reach end of life) or merge #651 for bootstrap5 support.

hodgestar:
I would say let's move forwards to bootstrap5. #651 approved.

PR #665, opened 2023-06-09
This setting is supposed to be a list/tuple, not a single string

drnlm:
See title

### DIFF
--- a/wafer/settings.py
+++ b/wafer/settings.py
@@ -202,7 +202,7 @@ AUTH_USER_MODEL = 'auth.User'
 REGISTRATION_FORM = 'wafer.registration.forms.WaferRegistrationForm'
 
 # Forms:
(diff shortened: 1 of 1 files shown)


wafer/settings.py
```diff
@@ -202,7 +202,7 @@
 REGISTRATION_FORM = 'wafer.registration.forms.WaferRegistrationForm'
 
 # Forms:
-CRISPY_ALLOWED_TEMPLATE_PACKS = "bootstrap5"
+CRISPY_ALLOWED_TEMPLATE_PACKS = ["bootstrap5", ]
```
hodgestar, line 205:
```suggestion
CRISPY_ALLOWED_TEMPLATE_PACKS = ["bootstrap5"]
```
We don't need the trailing comma for lists.

hodgestar (approved):

Issue #642, opened 2022-09-06
Include ticket type for registered speakers


drnlm:
We currently show whether the speaker is registered or not to admins when looking at the speaker profile - it would be helpful to also include information about the ticket type for cases (like PyConZA 2022) where there are different tickets for remote attendees.

drnlm:
This mostly now needs some UI work and some tests / handling for multiple tickets

PR #666, opened 2023-06-19
Feature/add ticket types to profile

drnlm:
To address #642 , this adds the tickets associated with a user to the profile.

This also tweaks the logic for associating tickets with a user so that different ticket types (conference, tutorial, etc) have fewer cases where the user will have to manually claim the ticket.

The actual UI part of the profile needs some love.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -140,6 +140,12 @@ Wafer's settings
     It should return a boolean result.
     The default function checks for a Quicket ticket.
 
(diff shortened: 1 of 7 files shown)


hodgestar (approved):
The changes look good to me.

I wondered a bit why the `WAFER_USER_TICKET_TYPES` setting exists, but I can imagine it being useful to reformat ticket names a bit.

PR #677, opened 2023-09-03
Feature/drop django before 3.2

drnlm:
While it is possible to run wafer with Django before 3.2, it requires a bunch of extra fiddling because several of our dependencies have dropped support for old django versions, including the defaults we use of crispy forms.

Given that Django 3.1 reach EOL in Dec 2021, and 2.2 in April 2022, there doesn't seem to be much value in keeping these as "officially supported"

This PR also adds 4.2 as a supported release, since there's nothing that changed from 4.1 that should break wafer, local testing seems fine and the tests all pass.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -27,7 +27,7 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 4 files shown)


hodgestar (approved):
Sounds good.

We aim to support the widest possible range of Django versions, but we can't realistically support versions for which support has been dropped in vital dependencies.

Issue #252, opened 2016-05-26
Allow public talk submissions


drnlm:
As requestted by Debconf people, so we don't lose track of it
- Expose all talk submissions, rather than just those accepted
  - Should this be totally public, or just to logged-in users?
  - Just logged in users reduces some of the spam concerns, but may not fit Debconf's requirements.
  - Do we need to restrict publishing by talk type?
- Possibly some sort of talk voting scheme?
  - Build this on top of the KeyValue stuff?

shirishag75:
As an interested person, I would suggest having some sort of privacy flip-flop built-in and letting the speakers/proposal makers/talk submitter also decide. Many a time people title their talk mysteriously and are gladly surprised. Having all the content before-hand would ruin the surprise for many, including myself (putting the attendee hat on) .

stefanor:
> Possibly some sort of talk voting scheme? 

Implemented in #418.

stefanor:
With some more thought, I think a useful refinement of this is:

There should be some talk types that are inherently public: e.g. BoF, ad-hoc session, open space submission. This can be a flag on the talk type.

There are spam concerns here, because it's making user-provided content public on the site. Events get spammy talk submissions (and user bios). So a useful additional feature here would be a "report content" button on any page with user-defined content. Obviously this can be a follow-up.

PR #683, opened 2023-10-08
Allow pending talks to be public, by TalkType

stefanor:
Add a new flag to TalkType to control whether pending talks are publicly visible.

Closes: #252


wafer/talks/views.py
```diff
@@ -52,7 +51,11 @@ def get_queryset(self):
         if self.request and Talk.can_view_all(self.request.user):
             talks = Talk.objects.all()
         else:
-            talks = Talk.objects.filter(Q(status=ACCEPTED) | Q(status=CANCELLED))
+            talks = Talk.objects.filter(
+                Q(status__in=(ACCEPTED, CANCELLED))
+                | Q(status__in=(SUBMITTED, UNDER_CONSIDERATION),
```
drnlm, line 56: Is there a reason this doesn't include Provisional as well?

drnlm (commented):
We have some documentation on the default workflow in docs/talks - that could use something about this new option

wafer/talks/views.py
stefanor: Good catch. Fixed.

drnlm (approved):

Issue #148, opened 2015-11-27
Make the UserProfile model customisable


madduck:
[This was originally just about removing the Twitter/Github fields for the DebConf context, but has since turned into a discussion about the UserProfile model in general]

======== original text ==========

Hell will break loose if we "condone" or "endorse" a non-free service such as Github or Twitter by even displaying form fields asking for these data.

Would it be possible to make their use optional? A good implementation of #147 (key-value store) will likely give those data a new home anyway, but meanwhile maybe a simple `settings.py` entry could be used to hide them from the web forms?

(I'll send the patch…)

drnlm:
I think this should be handled by having proper support for custom UserProfiles. Arguably wafer's UserProfile should not have these fields, and we should have a PyConUserProfile or DebconfUserProfile with the required information as required.

madduck:
Hi @drnlm, thanks for your thoughts. As I explained in #147, a conf-specific user profile just bears the danger of growing too large over time. Wafer's single-conference approach kinda guards against that, but I can still see much value in the KeyValueStore approach, especially if that storage became available through a REST API.

That said, and realistically speaking, we probably won't have this in time for DebConf16, which means we'll need a DebConfAttendeeProfile anyway.

Am I right in assuming that the main change to wafer required for this to happen would be to parametrise the actual UserProfile model used, e.g. to `settings.py`? So in our case, if setting `WAFER_PROFILE_MODEL=DebConfAttendeeProfile` (which derives from UserProfile) would cause Wafer to just use that instead everywhere, we'd be done, right?

And then we could monkey-patch the fields hinting at those evil proprietary services to oblivion and make it all fuzzy for us… ;)

stefanor:
You could probably also trivially hide these fields with CSS (hack hack hack)

madduck:
Good idea. I hope to have a look into overriding UserProfile within the next week or two, and will report back.

stefanor:
I think the entire user profile can be replaced with #147

madduck:
I am not entirely sure we *want* to do this. On the one hand, there is still the consideration of allowing multiple `UserProfile`s for an `auth.user` (e.g. for family, or colleagues).

On the other hand, at least when I envisioned the k-v-pairs solution, I thought of it as a datastore for per-conference, per-subteam data, sort of ad-hoc stuff, while stuff like email address, emergency contact, display name override, and the occasional other field are really "core" data.

[...]

PR #685, opened 2023-10-08
Feature/redo user profile with kv

### LINKED ISSUE #661: Ask for fediverse / mastodon id in profile page
Currently the profile asks for homepage, twitter and github. I think adding an option to publish fediverse / mastodon profile/id would be nice.

May be an easy way to add more fields down stream, like debconf23.debconf.org/ use wafer and it'd be good to be highlight username on salsa.debian.org there. When asked on the debconf team channel, it was suggested to ask upstream and doing that downstream is no easy.

drnlm:
This is a possible approach to getting rid of the hard coded twitter / github bits in the profile and making them more configurable.

We replace them with configured lists of sites (there's an argument we should have just one list), and use the key-value infrastructure to actually store the data.

I still need to look at validation and translation support.

This should resolve #661 and #148


wafer/users/forms.py
```diff
@@ -37,17 +39,62 @@ class Meta:
 class UserProfileForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        pre_social_index = len(self.fields)
+
+        # currently everything I'm asking for is an url, and this adds some
+        # validation - we may need to revisit this later
+        for field_name in settings.SOCIAL_MEDIA_ENTRIES:
+            self.fields[field_name] = forms.URLField(label=_(settings.SOCIAL_MEDIA_ENTRIES[field_name]),
```
stefanor, line 48: Using `_()` here won't extract the string into our `.po` files. Probably better to wrap it in settings.

wafer/users/models.py
```diff
@@ -44,11 +38,6 @@ class Meta:
     bio = models.TextField(_('bio'), null=True, blank=True)
 
     homepage = models.CharField(_('homepage'), max_length=256, null=True, blank=True)
-    # We should probably do social auth instead
-    # And care about other code hosting sites...
-    twitter_handle = models.CharField(_('Twitter handle'), max_length=15, null=True, blank=True,
-                                      validators=[TwitterValidator])
-    github_username = models.CharField(_('GitHub username'), max_length=32, null=True, blank=True)
```
stefanor, line 51: GitHub SSO uses this.

wafer/users/forms.py
```diff
@@ -37,17 +39,62 @@ class Meta:
...
-        self.helper['twitter_handle'].wrap(PrependedText,
-                                           '@', placeholder=_('handle'))
-        self.helper['github_username'].wrap(PrependedText,
-                                            '@', placeholder=_('username'))
+
+        # Add code hosting media header
+        # We do this in this order to avoid needing to do
+        # more maths
+        if settings.CODE_HOSTING_ENTRIES:
+            self.helper.layout.insert(pre_code_index, HTML('<p>Code</p>'))
```
stefanor, line 67: Wrap for translation?

stefanor (approved):
The styling is a bit basic, but that's never an issue here. :+1:

PR #686, opened 2023-10-17
Bugfix/fix invalid timestamp handling

drnlm:
Fixes a bug in how we handle invalid timestamps - previously timestamps that caused a ValueError would break things in interesting ways.


hodgestar (approved):
Whoops. Good catch.

PR #688, opened 2023-10-30
Feature/add speaker clash check

### LINKED ISSUE #680: Schedule validation check for speaker conflicts
Don't let one speaker speak in two venues at the same time. It's hard on the speakers.

We have a schedule validation framework, checking for this should be simple.

drnlm:
This adds a check that speakers aren't assigned to the same slot for multiple different things.

It also adds more tests for the other schedule validation functions, since that seems like a good idea

Closes #680

### DIFF
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -176,6 +176,9 @@ register_schedule_item_validator(
 register_schedule_item_validator(
         find_invalid_venues, 'venues',
         _('Invalid venues found in schedule.'))
+register_schedule_item_validator(
+        find_speaker_clashes, 'speaker_clashes',
+        _('Common speaker in simultaneous schedule items'))
 
 
 # Utility functions for checking the schedule state
--- a/wafer/schedule/admin.py
+++ b/wafer/schedule/admin.py
@@ -106,6 +106,29 @@ def find_invalid_venues(all_items):
     return venues.items()
 
 
+def find_speaker_clashes(all_items):
+    """Find items that have the same speaker and also have overlapping
+       slots"""
+    clashes = {}
+    seen_slots_speakers = {}
+    for item in all_items:
+        if item.talk:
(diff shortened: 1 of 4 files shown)


wafer/schedule/admin.py
```diff
@@ -106,6 +106,29 @@ def find_invalid_venues(all_items):
...
+            speakers = list(item.talk.authors.all())
+        elif item.page:
+            speakers = list(item.page.people.all())
+        for speaker in speakers:
+            for slot in item.slots.all():
+                candidate = (slot, speaker)
+                if candidate in seen_slots_speakers:
+                    if seen_slots_speakers[candidate] not in clashes:
+                        clashes[candidate] = [seen_slots_speakers[candidate]]
+                    clashes[candidate].append(item)
```
hodgestar, line 125: Is it possible to return which speaker has the clash? I guess in many cases it will be obvious but it might be easier for the user to sort out if we can display the extra information.

hodgestar (approved):
The new validator looks good and woot for the extra tests.

I left a suggestion for a further improvement, but also happy for this to be merged as is.

wafer/schedule/admin.py
drnlm: The error message already does include that, because it's built up from the returned items which includes those details (although the formatting could be improved)

```Common speaker in simultaneous schedule items: ((<Slot: Slot: Oct 14 (Fri): 10:00 - 11:00>, <User: neil>), [<ScheduleItem: T3est in Venue1 at Oct 14 (Fri): 10:00>, <ScheduleItem: tt in Venue2 at Oct 14 (Fri): 10:00>])```

drnlm: I've now improved the display in the admin page

hodgestar: Thanks!

PR #696, opened 2023-12-27
profile: add link to unregister if defined

terceiro:
This is used for DebConf and MiniDebConfs.


stefanor:
Seems like we should document these URLs. But yes, this does fit with the existing register URL.

PR #702, opened 2024-03-04
Feature/update ci

drnlm:
This updates the list of python & django versions we test against to the current set of supported releases.

It includes a fix so the tests pass on Django 5 (due to some changes in how admin filters are called in that version)


drnlm:
The failing selenium firefox tests is a timeout issue, and we should probably bump up the wait times a bit so we're not so sensitive if the runner is slow, but that should probably be a different PR.

hodgestar (approved):

PR #704, opened 2024-03-05
Feature/tweak selenium timing

drnlm:
This unifies the waits to use a common constant for easier tweaking, and adds a couple more wait points to hopefully make the selenium tests a bit less sensitive to the specific runner.

### DIFF
--- a/wafer/tests/utils.py
+++ b/wafer/tests/utils.py
@@ -52,6 +52,11 @@ def mock_avatar_url(self):
         return None
     return "avatar-%s" % self.user.email
 
+
+# Time to wait for elements to appear - single constant so we can easily tune it if needed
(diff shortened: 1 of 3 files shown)


wafer/tests/utils.py
```diff
@@ -52,6 +52,11 @@ def mock_avatar_url(self):
         return None
     return "avatar-%s" % self.user.email
 
+
+# Time to wait for elements to appear - single constant so we can easily tune it if needed
+WAIT_TIME = 20
```
hodgestar, line 57: I would give this a slightly more descriptive name like `SELENIUM_WAIT_TIME`.

hodgestar (approved):
One small suggestion but looks good to me.

PR #708, opened 2024-03-07
Fix Django 5 issue with the editor tests

drnlm:
Django 5's admin page is a bit different. and we need to up the selenium tests to account for that.

### DIFF
--- a/wafer/schedule/tests/test_schedule_editor.py
+++ b/wafer/schedule/tests/test_schedule_editor.py
@@ -17,7 +17,7 @@ except ImportError:
 
 from unittest import expectedFailure
 
-from django.utils import timezone
+from django.utils import timezone, version
 from django.urls import reverse
 
 from wafer.pages.models import Page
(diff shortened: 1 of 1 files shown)


wafer/schedule/tests/test_schedule_editor.py
```diff
@@ -133,10 +136,15 @@ def _start(self):
     def test_access_schedule_editor_no_login(self):
         """Test that the schedule editor isn't accessible if not logged in"""
         self.driver.get(self.edit_page)
-        header = WebDriverWait(self.driver, SELENIUM_WAIT_TIME).until(
-           expected_conditions.presence_of_element_located((By.TAG_NAME, "h1"))
-        )
-        self.assertEqual('Django administration', header.text)
+        if self.is_django_5:
```
hodgestar, line 139: Is it worthwhile wrapping these into a `wait_schedule_editor_ready()` method?

hodgestar (approved):
Looks good. Might be worthwhile wrapping the waiting incantation into a method but also happy for it to be merged as is.

wafer/schedule/tests/test_schedule_editor.py
drnlm: This construct is only used in the tests that aren't allowed to access the schedule editor, so it would be `wait_admin_error'` or something, and is currently only used in 2 of the tests.

it may be worth wrapping anyway, but it's not a huge win,and should possibly be done with a look at some of the other  tests to consolidate some other bits of repeated logic.

PR #725, opened 2024-10-06
Feature/more ticket info

drnlm:
This adds some more info, such as number of tickets sold per type and various filters, to the admin interface.

It also adds the ability to assign tags to ticket types - the main aim is to simplify constructing the checks like the `tickets_sold` used in the pyconza settings file (https://github.com/CTPUG/pyconza2024/blob/40c60bb3a005238ec666772c2969c86a82837f3e/settings.py#L156)  (count on tag rather than a list of ticket types)


wafer/tickets/models.py
```diff
@@ -2,15 +2,34 @@
 from django.conf import settings
 
 
+class TicketTypeTag(models.Model):
+    """Tags that can be added to a TicketType.
+
+       For example 'online, 'sponsor'"""
```
hodgestar, line 8:
```suggestion
    """Tags that can be added to a TicketType.

       For example 'online', 'sponsor'.
       """
```

hodgestar (changes requested):
Looks great. Thanks!

hodgestar (approved):

PR #729, opened 2024-11-25
talks: configure maximum number of talk authors to list explicitly

terceiro:
I have been using wafer instances for single-track events, where there is plenty of space in the schedule to just list all authors. This will let me configure my instances to list a higher number than 2.

### DIFF
--- a/wafer/talks/models.py
+++ b/wafer/talks/models.py
@@ -283,7 +283,7 @@ class Talk(models.Model):
             key=lambda author: u'' if author == self.corresponding_author
                                else author.userprofile.display_name())
         names = [author.userprofile.display_name() for author in authors]
-        if len(names) <= 2:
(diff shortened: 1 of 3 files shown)


wafer/talks/models.py
```diff
@@ -283,7 +283,7 @@ def get_authors_display_name(self):
             key=lambda author: u'' if author == self.corresponding_author
                                else author.userprofile.display_name())
         names = [author.userprofile.display_name() for author in authors]
-        if len(names) <= 2:
+        if len(names) <= settings.WAFER_SCHEDULE_MAX_AUTHORS:
             return u' & '.join(names)
```
drnlm, line 287: As a future improvement, it would be nice to tweak the formatting here for more than two names so it returns
Author A, Author B, Author C & Author D
instead of
Author A & Author B & Author C & Author D

drnlm (approved):

PR #732, opened 2025-03-17
Feature/update test matrix

drnlm:
Since more of our dependencies have dropped support for Django < 4.2, there seems little point in continuing to fight the march of progress. This PR updates the test matrix to just the currently supported Django set and also removes the pin on django-restframework we added when that dropped support for Django < 4.2.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -7,29 +7,27 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
-        django-version: ['3.2.0', '4.0.0', '4.1.0', '4.2.0', '5.0.0', 'main']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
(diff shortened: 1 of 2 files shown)


.github/workflows/django.yml
```diff
@@ -7,29 +7,27 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
-        django-version: ['3.2.0', '4.0.0', '4.1.0', '4.2.0', '5.0.0', 'main']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
```
hodgestar, line 10: We could also drop support for Python 3.8 since it's end of life now, although fine with keeping it if it isn't hurting us yet.

hodgestar (approved):

.github/workflows/django.yml
drnlm: Django 4.2 still officially supports Python 3.8, which is why I've left it in the test matrix.

PR #768, opened 2026-04-16
find_speaker_clashes: avoid crash without a talk or a page

terceiro:
If a user creates a schedule item and adds them to the schedule before associating a talk or a page, this causes a crash, because `speakers` will not be defined.

    UnboundLocalError: cannot access local variable 'speakers' where it
    is not associated with a value


terceiro:
The model validation should also probably disallow saving a schedule item without either a talk or a page.

drnlm (approved):

drnlm:
Thanks for the fix

Issue #615, opened 2021-08-01
Video flag for scheduled pages / bare schedule items


stefanor:
Currently in `pentabarf.xml`, the `<event>` for non-talk schedule items contains the static:

```xml
<recording>
    <optout>false</optout>
    <license>{{ WAFER_VIDEO_LICENSE }}</license>
</recording>
```

We should have some way to control this, so that video systems don't have to do heuristics.

Maybe a "no-video" boolean on scheduleitems, that overrides the default?

hodgestar:
Are ScheduleItems stable enough to attach this information to? I.e. would it be lost if an event was moved to a different slot?

stefanor:
Hrm, probably a duplicate of #384, which I didn't find in my quick skim earlier.

stefanor:
As you mentioned in #614, they probably aren't stable enough, but we could make them more stable.

PR #771, opened 2026-07-25
penta XML: Opt-out from video on pages by default

stefanor:
Typically we use pages for meals, and that means that they have always been marked as having video at DebConf.

Flip this to not having video. It's rare for something that's a page to have video coverage.

Obviously this is not perfect, it's replacing an assumption with a better one, see #615 and #384 for discussion on really improving this.

### DIFF
--- a/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
+++ b/wafer/schedule/templates/wafer.schedule/penta_schedule.xml
@@ -76,6 +76,7 @@
                             <license>{{ WAFER_VIDEO_LICENSE }}</license>
                           {% else %}
                             <optout>true</optout>
+                            <license/>
                           {% endif %}
                         </recording>
                       {% endwith %}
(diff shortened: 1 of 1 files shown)


drnlm (approved):

wafer/schedule/templates/wafer.schedule/penta_schedule.xml
```diff
@@ -107,8 +107,7 @@
                         <description/>
                       {% endif %}
                       <recording>
-                        <optout>false</optout>
-                        <license>{{ WAFER_VIDEO_LICENSE }}</license>
```
drnlm, line 111: Looks like we need to keep the license line to validate successfully.

stefanor: Filed https://github.com/frab/schedule.xml/issues/5 about that.

